PR #4, opened 2022-07-12
fix: use url module

mtfoley:

- [ ] 🍕 Feature
- [x] 🐛 Bug Fix
- [x] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI
- [ ] 📦 Chore (Release)
- [ ] ⏩ Revert

## Description

README update and different require for endpoint config. Just reflects what I needed to do in order to get this running locally.

## Added tests?

- [ ] 👍 yes

## Added to documentation?

- [ ] 📜 README.md
- [ ] 📓 docs.opensauced.pizza
- [ ] 🍕 dev.to/opensauced
- [ ] 📕 storybook

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -277,3 +277,4 @@ dist
 !.envrc
 **/supabase/.branches
 **/supabase/.temp
+data/*
--- a/README.md
+++ b/README.md
@@ -25,8 +25,8 @@
 ## 📖 Prerequisites
 
 In order to run the project we need the following software binaries installed on our development machines:
-- `node>=16.15.0`
-- `npm>=8.12.0`
+- `node>=18.0.0`
+- `npm>=8.12.1`
 - `docker>=20.10.12`
 - `supabase>=0.18.0`
 
--- a/README.md
+++ b/README.md
@@ -25,8 +25,8 @@
 ## 📖 Prerequisites
 
 In order to run the project we need the following software binaries installed on our development machines:
-- `node>=16.7.0`
-- `npm>=8.0.0`
+- `node>=16.15.0`
+- `npm>=8.12.0`
 - `docker>=20.10.12`
 - `supabase>=0.18.0`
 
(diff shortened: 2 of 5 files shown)


mtfoley:
@0-vortex I've been doing some other poking around and found that if I seed locally from what's in the open-sauced/actions/hot-supabase directory, I have to drop the repos/users foreign key constraint.

Also, when I do `supabase db reset` and the `seed.sql` file gets used, the data doesn't make it into the tables and I wonder if something similar is going on.

0-vortex:
> @0-vortex I've been doing some other poking around and found that if I seed locally from what's in the open-sauced/actions/hot-supabase directory, I have to drop the repos/users foreign key constraint.
> 
> 
> 
> Also, when I do `supabase db reset` and the `seed.sql` file gets used, the data doesn't make it into the tables and I wonder if something similar is going on.

are you running supabase locally? otherwise you are juggling between adjusting the dev or the prod db, both things which shouldn't do right now as the entire structure should sit in the backend rather than actions.

To figure out if the seeds are broken just run them manually in supabase ui 😁

mtfoley:
Yeah, I'm running it locally. i saw the docker containers pull and spin up, and i saw the `supabase/seed.sql` file get used. however, in the studio, the tables are empty. i realize now, the shell script to download copies of the SQL files from the actions repo was unnecessary, but i still wonder if we should troubleshoot why that initial seed file didn't fill up the db locally.

0-vortex:
> Yeah, I'm running it locally. i saw the docker containers pull and spin up, and i saw the `supabase/seed.sql` file get used. however, in the studio, the tables are empty. i realize now, the shell script to download copies of the SQL files from the actions repo was unnecessary, but i still wonder if we should troubleshoot why that initial seed file didn't fill up the db locally.

Sorry for the late reply - the seed file is being executed via psql inside the docker containers, similarly to what we are doing in the seed.sh script, hence the errors are not visible unless you check the process logs after you sh into the containers.
You can check the error yourself simply by renaming seed.sql and running it locally as described above, however the most likely culprit is bad insert order, conflicting with relationships, running the individual seed files in the logical order would fix this (i normally seed it this way and then adjust script in actions, however we are partially divergent until backend is deployed fully)

0-vortex:
We are moving off of this system for seeding, will close this :D

PR #31, opened 2022-09-16
feat: add endpoint to retrieve onboarding repos based on the PAT

brandonroberts:

- [x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI
- [ ] 📦 Chore (Release)
- [ ] ⏩ Revert

## Description

Pulls the available repos for a given Personal Access Token as an array of strings

## Added tests?

- [ ] 👍 yes

## Added to documentation?

- [ ] 📜 README.md
- [ ] 📓 docs.opensauced.pizza
- [ ] 🍕 dev.to/opensauced
- [ ] 📕 storybook

### DIFF
--- a/src/auth/auth.controller.ts
+++ b/src/auth/auth.controller.ts
@@ -1,5 +1,6 @@
-import { Controller, Get, HttpCode, HttpStatus, Post, UseGuards } from "@nestjs/common";
+import { Body, Controller, Get, HttpCode, HttpStatus, NotFoundException, Post, UseGuards } from "@nestjs/common";
 import { ApiBearerAuth, ApiOkResponse, ApiOperation, ApiTags } from "@nestjs/swagger";
+import { execSync } from "child_process";
 import { SupabaseGuard } from "./supabase.guard";
 import { SupabaseAuthUser } from "nestjs-supabase-auth";
 import { User } from "./supabase.user.decorator";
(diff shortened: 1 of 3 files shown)


src/auth/auth.controller.ts
```diff
@@ -71,4 +72,32 @@ export class AuthController {
...
+    @Body() body: string,
+  ): string[] {
+    const { token } = JSON.parse(body || "{}") as { token: string | undefined };
+
+    if (!token) {
+      throw (new NotFoundException);
+    }
+
+    try {
+      const repoList = execSync(`npx repo-report ls --token="${token}"`).toString("utf-8");
```
0-vortex, line 95: We can't spawn on droplets but you can use the `getRepositories` function with normal await - it's just gonna do a graphql request :D

src/auth/auth.controller.ts
```diff
@@ -71,4 +72,32 @@ export class AuthController {
...
+  @ApiOperation({
+    operationId: "postOnboardingRepos",
+    summary: "Retrieves onboarding repos using a PAT",
+  })
+  @ApiOkResponse()
+  @HttpCode(HttpStatus.OK)
+  postOnboardingRepos (
+    @Body() body: string,
+  ): string[] {
+    const { token } = JSON.parse(body || "{}") as { token: string | undefined };
```
0-vortex, line 88: I think you need the `@User() user: SupabaseAuthUser` decorator here

src/auth/auth.controller.ts
```diff
@@ -71,4 +72,32 @@ export class AuthController {
...
+
+    if (!token) {
```
0-vortex, line 90: this block is redundant

0-vortex (changes requested):
Some minor changes here and there

src/auth/auth.controller.ts
brandonroberts: Is this so it will force an authenticated session to be provided?

0-vortex: Yes, we can't let users ddos this API out by providing a custom broken tokens, but we can store a token encrypted for the logged in user and then re-use it

PR #77, opened 2023-01-09
feat: add endpoint to find a user by username

brandonroberts:

- [x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI
- [ ] 📦 Chore (Release)
- [ ] ⏩ Revert

## Description

This PR:
- Adds the `/users/:username` endpoint to find a user by username

## Added tests?

- [ ] 👍 yes

## Added to documentation?

- [ ] 📜 README.md
- [ ] 📓 docs.opensauced.pizza
- [ ] 🍕 dev.to/opensauced
- [ ] 📕 storybook

### DIFF
--- a/src/user/user.service.ts
+++ b/src/user/user.service.ts
@@ -44,7 +44,7 @@ export class UserService {
 
     queryBuilder.where("LOWER(login) = LOWER(:username)", { username });
 
-    let item: DbUser | null = await queryBuilder.getOne();
+    const item: DbUser | null = await queryBuilder.getOne();
 
     if (!item) {
       throw (new NotFoundException);
--- a/src/user/user.service.ts
+++ b/src/user/user.service.ts
@@ -44,14 +44,7 @@ export class UserService {
 
     queryBuilder.where("LOWER(login) = LOWER(:username)", { username });
 
-    let item: DbUser | null;
-
-    try {
-      item = await queryBuilder.getOne();
-    } catch (e) {
-      // handle error
-      item = null;
-    }
+    let item: DbUser | null = await queryBuilder.getOne();
 
     if (!item) {
       throw (new NotFoundException);
--- a/src/user/user.service.ts
+++ b/src/user/user.service.ts
@@ -39,6 +39,27 @@ export class UserService {
     return item;
   }
 
+  async findOneByUsername (username: string): Promise<DbUser> {
+    const queryBuilder = this.baseQueryBuilder();
+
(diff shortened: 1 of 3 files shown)


src/user/user.service.ts
```diff
@@ -39,6 +39,27 @@ export class UserService {
...
 
+  async findOneByUsername (username: string): Promise<DbUser> {
+    const queryBuilder = this.baseQueryBuilder();
+
+    queryBuilder.where("LOWER(login) = LOWER(:username)", { username });
+
+    let item: DbUser | null;
+
+    try {
+      item = await queryBuilder.getOne();
```
0-vortex, line 50: Are you sure you need the try/catch here? 😅 

You're not really doing anything with the error, this is like hiding it from our system

0-vortex (changes requested):
Not sure about empty errors, would be best to just leave the default behavior 👍

src/user/user.service.ts
brandonroberts: ✂️

src/user/user.service.ts
```diff
@@ -39,6 +39,20 @@ export class UserService {
     return item;
   }
 
+  async findOneByUsername (username: string): Promise<DbUser> {
+    const queryBuilder = this.baseQueryBuilder();
+
+    queryBuilder.where("LOWER(login) = LOWER(:username)", { username });
+
+    const item: DbUser | null = await queryBuilder.getOne();
```
0-vortex, line 47: `getOneOrFail` maybe?

brandonroberts: We aren't using this anywhere else, will the error still be presented the same way?

brandonroberts: This throws a 500, where we need a try/catch

0-vortex: ok with throwing a new one on top of https://docs.nestjs.com/exception-filters#built-in-http-exceptions

brandonroberts: So catch and rethrow `NotFoundException`?

0-vortex: no, like the first time you did, when you threw on null and let all the other errors propagate before that

0-vortex: or well, second :D

PR #119, opened 2023-03-09
feat: update pr search endpoint and filters to use pull_requests_insights

brandonroberts:

- [x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI
- [ ] 📦 Chore (Release)
- [ ] ⏩ Revert

## Description

This PR:
- Updates pull requests search endpoint to use pull_requests_insights table
- Updates queries and filters to remove joins
- Adds pull_requests_insights table to migrations

## Added tests?

- [ ] 👍 yes

## Added to documentation?

- [ ] 📜 README.md
- [ ] 📓 docs.opensauced.pizza
- [ ] 🍕 dev.to/opensauced
- [ ] 📕 storybook

### DIFF
--- a/migrations/20230309134600_add_pull_requests_insights.sql
+++ b/migrations/20230309134600_add_pull_requests_insights.sql
@@ -1,0 +1,59 @@
+create table if not exists public.pull_requests_insights
+(
+	id integer,
+	repo_id integer,
+	number integer,
+	state text,
+	draft boolean,
+	title text,
+	source_label text default ''::text,
+	target_label text default ''::text,
+	target_branch text,
+	source_branch text,
+	author_login text,
+	author_avatar text,
+	assignee_login text default ''::text,
+	assignee_avatar text default ''::text,
+	assignees_login text[],
+	assignees_avatar text[],
+	requested_reviewers_login text[],
+	label_names text[],
+	labels jsonb,
+	merged boolean default false,
+	mergeable boolean default false,
+	rebaseable boolean default false,
+	created_at timestamp default now() not null,
+	merged_at timestamp,
+	updated_at timestamp default now() not null,
+	last_updated_at timestamp default now() not null,
+	closed_at timestamp,
+	additions integer default 0,
+	deletions integer default 0,
+	changed_files integer default 0,
+	comments integer default 0,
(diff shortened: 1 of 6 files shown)


0-vortex:
Does this also need a timeframe? like snapshot day?

brandonroberts:
@0-vortex not sure. The plan is to remove/re-insert the records on an interval

0-vortex:
> @0-vortex not sure. The plan is to remove/re-insert the records on an interval. We're already querying by the date range

I checked again and this is not really flattening anything, this is just moving to a smaller subset, making the scans manageable via the etl cron script - it would be best to add a "day" type field and only query the "latest day". This would give us exactly 1000 x {noOfFilters} records per day, which would represent every user shown in the chart.

The aggregation would represent values for the actual date filter used now, 7, 30 and 90 days. We can't go back in time or trust the data if we keep rewriting it 😅

brandonroberts:
The 1000 count is not fixed because we show data on the graph based on a dynamic combination of multiple filters. 

Filters
 - Day range (last 30 days, fixed)
 - Topic (javascript, rust, etc)
 - State (open, closed)
 - Contributor (specific)
 - Repositories (1 to many)
 - Status (Recent, Top 100, Has Spam, Most Active, Minimum Contributors)

0-vortex:
> The 1000 count is not fixed because we show data on the graph based on a dynamic combination of multiple filters.
> 
> Filters
> 
> * Day range (last 30 days, fixed)
> * Topic (javascript, rust, etc)
> * State (open, closed)
> * Contributor (specific)
> * Repositories (1 to many)
> * Status (Recent, Top 100, Has Spam, Most Active, Minimum Contributors)

`(each topic) x (each state) x (each status)` flattened in groups, get `limit X` will result in actual X users end result, the ETL script flattening this would have to go through all the combinations; then the graph displays latest contributions (pull requests?) from those users, based on the same above filters, so it would IMHO definitely have to loop twice in the ETL to compute everything 😅

Issue #215, opened 2023-07-20
Feature: Test user services  and controllers


takanome-dev:
### Type of feature

✅ Test

### Current behavior

Issue tracking for services in the `user` folder/entity
- [ ] Test user service/controller
- [ ] Test user collaboration service/controller
- [ ] Test user follows service/controller
- [ ] Test user highlights service/controller
- [ ] Test user notification service/controller

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

### Contributing Docs

- [ ] I agree to follow this project's Contribution Docs

PR #219, opened 2023-07-20
test: add unit test for user service

takanome-dev:
## Description
This PR adds unit tests for the user service. The unit tests ensure the functionality of the user service is properly tested and validated.

_Generated using [OpenSauced](https://opensauced.ai/)._

- [ ] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [x] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI
- [ ] 📦 Chore (Release)
- [ ] ⏩ Revert

## Added tests?

- [ ] 👍 yes

## Added to documentation?

- [ ] 📜 README.md
- [ ] 📓 docs.opensauced.pizza
- [ ] 🍕 dev.to/opensauced
- [ ] 📕 storybook

### DIFF
--- a/src/user/services/user/user.service.spec.ts
+++ b/src/user/services/user/user.service.spec.ts
@@ -1,0 +1,151 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
+import { NotFoundException } from "@nestjs/common";
+import { getRepositoryToken } from "@nestjs/typeorm";
+import { Test } from "@nestjs/testing";
+import { faker } from "@faker-js/faker";
+import { ObjectLiteral, Repository } from "typeorm";
+
+import { UserService } from "../../services/user/user.service";
+import { userNotificationTypes } from "./../../entities/user-notification.constants";
+import { DbUser } from "../../user.entity";
+import { DbUserHighlightReaction } from "../../entities/user-highlight-reaction.entity";
+
+type MockRepository<T extends ObjectLiteral = any> = Partial<Record<keyof Repository<T>, jest.Mock>>;
+const createMockRepository = <T extends ObjectLiteral = any>(): MockRepository<T> => ({
+  createQueryBuilder: jest.fn(),
+});
+
+describe("UserService", () => {
+  let userService: UserService;
+  let dbUserRepositoryMock: MockRepository;
+  let dbUserHighlightReactionRepositoryMock: MockRepository;
+
+  beforeEach(async () => {
+    const module = await Test.createTestingModule({
+      providers: [
+        UserService,
+        {
+          provide: getRepositoryToken(DbUser, "ApiConnection"),
+          useValue: createMockRepository(),
+        },
+        {
+          provide: getRepositoryToken(DbUserHighlightReaction, "ApiConnection"),
+          useValue: createMockRepository(),
+        },
+      ],
+    }).compile();
+
+    userService = module.get<UserService>(UserService);
+    dbUserRepositoryMock = module.get<MockRepository>(getRepositoryToken(DbUser, "ApiConnection"));
+    dbUserHighlightReactionRepositoryMock = module.get<MockRepository>(
+      getRepositoryToken(DbUserHighlightReaction, "ApiConnection")
+    );
+  });
+
+  it("should be defined", () => {
+    expect(userService).toBeDefined();
+  });
+
+  describe("findTopUsers", () => {
+    it("should return a list of top users", async () => {
+      const expectedUsers = [{ id: faker.number.int() }, { id: faker.number.int() }];
+      const defaultLimit = 10;
+
+      const createQueryBuilderMock = {
+        select: jest.fn().mockReturnThis(),
+        innerJoin: jest.fn().mockReturnThis(),
+        where: jest.fn().mockReturnThis(),
+        groupBy: jest.fn().mockReturnThis(),
+        orderBy: jest.fn().mockReturnThis(),
+        limit: jest.fn().mockReturnThis(),
+        getRawMany: jest.fn().mockResolvedValue(expectedUsers),
+      };
+
+      dbUserHighlightReactionRepositoryMock.createQueryBuilder?.mockReturnValue(createQueryBuilderMock);
+      const result = await userService.findTopUsers(defaultLimit);
+
+      expect(dbUserHighlightReactionRepositoryMock.createQueryBuilder).toHaveBeenCalled();
+      /*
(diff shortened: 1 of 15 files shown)


src/user/services/user/user.service.spec.ts
```diff
@@ -0,0 +1,239 @@
...
+        is_open_sauced_member: true,
+        login: supabaseUser.user_metadata.user_name,
+        email: supabaseUser.user_metadata.email,
+        created_at: new Date(supabaseUser.identities[0].created_at),
+        updated_at: new Date(supabaseUser.identities[0].updated_at),
+        connected_at: new Date(supabaseUser.confirmed_at),
+      };
+
+      expect(dbUserRepositoryMock.create).toHaveBeenCalledWith(newUser);
+      // ! need help here, I don't know how to test the `save` method in the `checkAddUser` method (newUser.save()) 🙃
```
takanome-dev, line 227: cc @brandonroberts @AlexHladin 

I tried `expect(dbUserRepositoryMock.save).toHaveBeenCalled()` but it didn't work because it's not the `userRepository.save()` that has been called but `newUser.save()` which is an instance of that class

src/user/services/user/user.service.spec.ts
```diff
@@ -0,0 +1,239 @@
...
+  describe("findOneById", () => {
+    const user = { id: faker.number.int(), email: faker.internet.email() };
+    const createQueryBuilderMock = {
+      addSelect: jest.fn().mockReturnThis(),
+      where: jest.fn().mockReturnThis(),
+      setParameters: jest.fn().mockReturnThis(),
+      getOne: jest.fn().mockResolvedValue(user),
+    };
+
+    /**
```
takanome-dev, line 92: I removed the test suite when `includeEmail` is false because it always includes the email. IMO it's because I mocked the response. Maybe this should be tackled when doing e2e

src/user/services/user/user.service.spec.ts
```diff
@@ -0,0 +1,239 @@
...
+     * @author @takanome-dev
+     * @todo test with a real db when doing e2e if `findOneById`
+     * with includeEmail=false is returning the email or not.
+     * Cannot be tested with unit test (IMO) since the result is mocked.
+     */
```
takanome-dev, line 97: I removed the test suite when `includeEmail` is false because it always includes the email. IMO it's because I mocked the response. Maybe this should be tackled when doing e2e

src/user/services/user/user.service.spec.ts
AlexHladin: @takanome-dev you can do a simple thing:
```typescript
const saveMock = jest.fn()
dbUserRepositoryMock.create.mockReturnValue(createMock)
...
expect(saveMock).toHaveBeenCalled()
```

takanome-dev:
@brandonroberts I didn't add tests for the controller because it's just the same mocks and tests as the service (just duplicates). It would be better to test them when doing e2e.

That's just my opinion, do let me know if you think I should do it.

brandonroberts:
@takanome-dev seems reasonable to me

src/user/services/user/user.service.ts
```diff
@@ -3,27 +3,25 @@ import { Repository, SelectQueryBuilder } from "typeorm";
...
+import { PageMetaDto } from "../../../common/dtos/page-meta.dto";
 
 @Injectable()
 export class UserService {
   constructor(
     @InjectRepository(DbUser, "ApiConnection")
     private userRepository: Repository<DbUser>,
     @InjectRepository(DbUserHighlightReaction, "ApiConnection")
-    private userHighlightReactionRepository: Repository<DbUserHighlightReaction>,
-    private pagerService: PagerService
```
jpmcb, line 26: I'm guessing the `pagerService` was unused in this service?

jpmcb:
Is there a related tracking issue or ticket for this? If not, let's make one so we can track testing efforts across the API.

> Note that user.service has been moved to a new folder services/user for a better structure. The other services will be moved in future prs.

I'd like to understand this more: what benefit do we get from moving the services under a `services/` directory? Better structure in the code? Is this a nest js convention?

src/user/services/user/user.service.ts
takanome-dev: Yeah, but it should be used to remove duplicates. We should open an issue about cleaning up the usage of `new PageMetaDto()` and `new PageDto()` in services.

takanome-dev:
> Is there a related tracking issue or ticket for this? If not, let's make one so we can track testing efforts across the API.
> 
Yeah -> https://github.com/open-sauced/api/issues/215

> I'd like to understand this more: what benefit do we get from moving the services under a `services/` directory? Better structure in the code? Is this a nest js convention?
>
Just for better structure. There are 4 other services that need to be tested. Dumping all the test files in the user module may be messy IMO.

brandonroberts:
I don't think we need to nest the folders too much. There's already a `user` folder, no need to have a `services/user` folder with one service inside there.

If we have `user/services` and all the services are in there that should be good enough. We don't want to create a complex folder structure that's too granular.

takanome-dev:
> If we have `user/services` and all the services are in there that should be good enough. We don't want to create a complex folder structure that's too granular.

@brandonroberts done :+1:

src/user/services/user.service.ts
```diff
@@ -176,7 +174,7 @@ export class UserService {
       return user;
     } catch (e) {
       // create new user
-      const newUser = this.userRepository.create({
+      const newUser = await this.userRepository.save({
```
brandonroberts, line 177: Revert this

brandonroberts (approved):
LGTM. Revert the user service and open a separate PR unless it impacts the tests

takanome-dev:
> LGTM. Revert the user service and open a separate PR unless it impacts the tests

Yeah it does, as mentioned [here](https://github.com/open-sauced/api/pull/219#discussion_r1270478390) it was failing the test and I couldn't figure out a way to mock that instance.

brandonroberts:
@takanome-dev got it. Going forward, we shouldn't have to modify existing working code to make tests pass

takanome-dev:
> @takanome-dev got it. Going forward, we shouldn't have to modify existing working code to make tests pass

Sure, will keep that in mind. Thanks :+1:

OgDev-01 (approved):

Issue #218, opened 2023-07-20
Feature: integrate `pizza` service with API


jpmcb:
### Type of feature

🍕 Feature

### Current behavior

Currently, [the `pizza` oven service](https://github.com/open-sauced/pizza) (which indexes and processes commits on repositories) is not integrated in the API.

Ideally, we want a _single_ API layer for all things OpenSauced.

### Suggested solution

Let's integrate the pizza oven service into the API. Initially, I'd imagine a few routes:

Get commits for a repository:
```
/v1/repos/{id}/commits
```

Bake a repository (which will send the request to the Pizza oven service to then do all the indexing):
```
/v1/repos/{id}/bake
```

This involves a few things:
- Aligning the database design with what the `pizza` oven service (cc @brandonroberts) so that the API can also consume the indexing from the pizza oven through to the API
- Align the above 2 initial routes (does it make sense that these would live on the `/v1/repos` endpoint?

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

### Contributing Docs

- [X] I agree to follow this project's Contribution Docs

jpmcb:
Chatted with @brandonroberts this morning, and we're thinking we'll need to make a few changes:
- Create 3 new tables on the database that align with the `pizza-oven` service's data (which is only processing commits, not git service provider data):
    - `baked-repos`: A table of ids and urls that have been processed by the pizza-oven service
        - This will act as the sort of "bridge" table to link the existing GitHub data in the API and the data the pizza oven service generates
    - `commit-authors`: A table of ids and emails associated with commits
    - `commits`: Which has a foreign key to the `baked-repos` and `commit-authors`. This table contains all the commit data.
    - Then, the `pizza-oven` service can be aligned to write data into the API's postgres database.

- `/v1/repos/{id}/commits` can be used to get the commits for an individual GitHub repo. This may also involve abit of work getting the two tables to "link" but might just be an SQL query.

- `/v1/bake/{git-url}` would probably be a better endpoint the CLI and other clients can use to forward post requests to the `pizza-oven` service. This way, it's not linked to the ID of a GitHub entry in the database.

bdougie:
Related extra information. 

While working last night noticed the term mannequins. Contributions without a github user. Thinking about the duplication work we will need to do is similar. While we list commits, we might need to identify commits that may be stood up by a mannequin

[...]

Issue #11, opened 2022-08-02
Feature: implement cryptographic functions


0-vortex:
### Type of feature

🍕 Feature

### Current behavior

Storing sensitive user data like PATs can be done securely by encrypting this data in our storage tables with a quantum resistant crypto algorithm and storing the key outside of this repository or user code.

In combination with the supabase JWT secret leaves no room for exploitation outside of a full breach! :rock

### Suggested solution

Find quantum resistant crypto algos, ideally a javascript implementation available via `npm` and create and environment variable to store the secret.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

### Contributing Docs

- [X] I agree to follow this project's Contribution Docs

PR #225, opened 2023-07-26
feat: integrates pizza oven service with open-sauced API

jpmcb:
## Description

- adds 3 tables (including entities / dtos) so that the pizza oven service can integrate directly with the open sauced API and its data model.
- Adds `bake/` route:
  - POST `/v1/bake` - send a post request to the pizza oven micro-service using the git clone url
  - GET `/v1/bake/repos/{id}` - get a baked repo by id
  - GET `/v1/bake/repos/list` - get a list of baked repos
  - GET `/v1/bake/commit-authors/{id}` - get a commit author by their id
  - GET `/v1/bake/commit-authors/list` - list and paginate all commit authors
  - GET `/v1/bake/commits/{id}` - get a commit by its id
  - GET `/v1/bake/commits/list/repo/:id` - get a paginated list of commits by a baked repo's id
  - GET `/v1/bake/commits/list/commit-authors/:id` - get a paginated list of commits by a commit author's id

- [x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI
- [ ] 📦 Chore (Release)
- [ ] ⏩ Revert

Related to #218 - once we've arrived on consensus for this data model, this will require a followup in the `pizza` repo to update it's model to integrate with the APIs database.

A followup PR in the `pizza` service will fully integrate this with this data model: https://github.com/open-sauced/pizza/pull/30

N/a

## Added tests?

- [ ] 👍 yes
- [x] 🙋 no, because I need help

## Added to documentation?

- [ ] 📜 README.md
- [ ] 📓 docs.opensauced.pizza
- [ ] 🍕 dev.to/opensauced
- [ ] 📕 storybook

There are some deployment steps that will need to take place:
- [ ] Apply the migration to our prod / beta dbs
- [ ] integrate pizza oven microservice with the beta / production dbs
- [ ] integrate the pizza CLI with this new API (to be able to post requests)
- [ ] Deploy this new API


jpmcb:
@brandonroberts - let me know if this looks like the right track: I think this data model will work well for us and let us stitch the `commits` and `repos` tables together nicely.

brandonroberts:
@jpmcb you're on the right track from what I see so far. I'll take a further look also. We don't need to guard against duplicate repos being baked here correct? That will be done at the pizza oven service level?

jpmcb:
> We don't need to guard against duplicate repos being baked here correct? That will be done at the pizza oven service level?

Correct. This integration is really twofold: a thin wrapper to send POST requests to the configured pizza oven service and a querier to get commits and things out of the database. Those post requests are forwarded on with little validation. The `pizza` oven microservice handles that side of it all and will connect to the database to add new commits / repos.

migrations/20230726151100_add_pizza_service_tables.sql
```diff
@@ -0,0 +1,54 @@
+------------------------------------
+-- Pizza oven "baked" repos table --
+-------------------------------------
+create table if not exists public.baked_repos
```
brandonroberts, line 4: We could consider adding a `repo_id` as a foreign key for the `repos` table. It may provide an easy indicator of which baked repos that have not been indexed yet.

migrations/20230726151100_add_pizza_service_tables.sql
```diff
@@ -0,0 +1,54 @@
...
+--------------------------------
+-- Pizza oven indexed commits --
+--------------------------------
+
+create table if not exists public.commits (
+  id bigint not null generated by default as identity ( increment 1 start 1 minvalue 1 maxvalue 9223372036854775807 cache 1 ),
+  baked_repo_id bigint not null references public.baked_repos (id) on delete cascade on update cascade,
+  commit_author_id bigint not null references public.commit_authors(id) on delete cascade on update cascade,
+  commit_hash character varying(255) collate pg_catalog."default" not null,
+  commit_date timestamp with time zone not null default now(),
```
brandonroberts, line 45: We usually store timestamp without the timezone, so when its displayed to the user, there local timezone is used.

```suggestion
  commit_date timestamp without time zone not null default now(),
```

src/pizza/commit-authors.service.ts
```diff
@@ -0,0 +1,49 @@
...
+    const queryBuilder = this.baseQueryBuilder();
+
+    queryBuilder.offset(pageOptionsDto.skip).limit(pageOptionsDto.limit);
+
+    const itemCount = await queryBuilder.getCount();
+    const entities = await queryBuilder.getMany();
+
+    const pageMetaDto = new PageMetaDto({ itemCount, pageOptionsDto });
+
+    return new PageDto(entities, pageMetaDto);
```
AlexHladin, line 33: @jpmcb you can use [Pager service](https://github.com/open-sauced/api/blob/beta/src/common/services/pager.service.ts) to avoid duplication of code for pagination.

src/pizza/commits.service.ts
```diff
@@ -0,0 +1,66 @@
...
+    queryBuilder.where(`baked_repo_id = ${id}`);
+
+    queryBuilder.offset(pageOptionsDto.skip).limit(pageOptionsDto.limit);
+
+    const itemCount = await queryBuilder.getCount();
+    const entities = await queryBuilder.getMany();
+
+    const pageMetaDto = new PageMetaDto({ itemCount, pageOptionsDto });
+
+    return new PageDto(entities, pageMetaDto);
```
AlexHladin, line 35: @jpmcb the same here

src/pizza/commits.service.ts
```diff
@@ -0,0 +1,66 @@
...
+  baseQueryBuilder() {
+    const builder = this.commitsRepository.createQueryBuilder("commits");
+
+    return builder;
+  }
+
+  async findAllCommitsByBakedRepoId(pageOptionsDto: PageOptionsDto, id: number): Promise<PageDto<DbCommits>> {
+    const queryBuilder = this.baseQueryBuilder();
+
+    queryBuilder.where(`baked_repo_id = ${id}`);
```
brandonroberts, line 26: TypeORM will sanitize the input if we use parameterized queries.

```suggestion
    queryBuilder.where(`baked_repo_id = :id`, { id });
```

src/pizza/commits.service.ts
```diff
@@ -0,0 +1,66 @@
...
+
+    const pageMetaDto = new PageMetaDto({ itemCount, pageOptionsDto });
+
+    return new PageDto(entities, pageMetaDto);
+  }
+
+  async findAllCommitsByCommitAuthorId(pageOptionsDto: PageOptionsDto, id: number): Promise<PageDto<DbCommits>> {
+    const queryBuilder = this.baseQueryBuilder();
+
+    queryBuilder.where(`commit_authors_id = ${id}`);
```
brandonroberts, line 41: Same with parameterized queries

src/pizza/pizza-oven.service.ts
```diff
@@ -0,0 +1,67 @@
+import { Injectable, NotFoundException } from "@nestjs/common";
+import { InjectRepository } from "@nestjs/typeorm";
+
+import { DbBakedRepo } from "./entities/baked-repo.entity";
```
brandonroberts, line 4: Let's group the third-party imports at the top, followed by our imports last. I think there's a lint rule we can enable to fix this automatically

src/pizza/pizza-oven.service.ts
```diff
@@ -0,0 +1,67 @@
...
+    const host: string = this.configService.get("pizza.host")!;
+    const port: string = this.configService.get("pizza.port")!;
+
+    return lastValueFrom(this.httpService.post(`${host}:${port}/bake`, data).pipe(map((resp) => resp.status)));
+  }
+
+  async findBakedRepoById(id: number): Promise<DbBakedRepo> {
+    const queryBuilder = this.baseQueryBuilder();
+
+    queryBuilder.where(`id = ${id}`);
```
brandonroberts, line 44:
```suggestion
    queryBuilder.where(`id = :id`, { id });
```

brandonroberts (changes requested):

migrations/20230726151100_add_pizza_service_tables.sql
jpmcb: I had originally considered using the date/time directly from the commits without the timestamp in the database, but that created some really weird behavior in the pizza service when attempting to get the "latest" commits by date. The wrong commit would be returned since commits are timestamped by the individual commit author (which may have originated from within any timezone),

Attempting to drop them into the database _without_ a timestamp results in the unexpected ordering by date. So, my solution in the pizza service was the convert them during indexing to UTC:

https://github.com/open-sauced/pizza/blob/e99d988bbc30d3fc9fa4e4a6cbb0d789a5298d76/pkg/server/server.go#L169

and enter them in the database with the UTC timestamp explicitly.

migrations/20230726151100_add_pizza_service_tables.sql
jpmcb: Sounds good. That'll probably require a change on the `pizza` side as well to query for a `repo` that has the same `clone_url`

brandonroberts: Ok. If you think using the `clone_url` as the check, I think its fine. Its more so we can check if the repo is index, and also join on this table without using the `clone_url`

bdougie (commented):
Storing public commit emails makes me nervous. I don't know if bcrypt is the answer but I'd love to avoid a [Geekedin situation](https://www.troyhunt.com/8-million-github-profiles-were-leaked-from-geekedins-mongodb-heres-how-to-see-yours/).

src/pizza/dtos/commit_authors.dto.ts
```diff
@@ -0,0 +1,12 @@
+import { ApiProperty } from "@nestjs/swagger";
+import { IsString } from "class-validator";
+
+export class CommitAuthorDto {
+  @ApiProperty({
+    description: "Public Commit Author Email",
+    type: String,
+    example: "hello@opensauced.pizza",
```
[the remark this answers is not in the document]
jpmcb: Totally: I think that's wise. We'll want to somehow store those emails in a secure manner so that we can get commit data from users who eventually opt-in with their "known aliases".

Is there any precedence for this currently in our API? Or is this somewhat blocked on #11?

brandonroberts:
With that concern, I'd recommend not exposing any endpoints for commit authors, and make the commit author email hidden by default by adding a `select: false` to the entity field. 

https://github.com/open-sauced/api/pull/225/files#diff-9a7453839155c8f1da849c06933fa0882375314495c99312e0ae7fc8e8915ce5R21

That way its insert-only. We will need to revisit how we use insights based on this data also. 

On the encryption side, we don't have anything existing there, but I'm sure there's a javascript package our there for bcrypt or whatever we choose to allow us to encrypt the email before inserting the record.

brandonroberts (changes requested):

jpmcb:
Revisiting this and pushed changes that address:

- Use the `"id = :id", { id }` for sanitizing selecting on id.
- Makes the commit authors field hidden per @brandonroberts's suggestion.
  - Does punting on the encryption bits make sense right now? If this field is hidden via our API, that should at least keep people from scraping emails. But we may want to investigate adding encryption at rest for other services that may read this data from the database.
- Adds a `repo_id` field on the baked repos table that references an id in the repos table.
  - I'm thinking that we'll have the ETL service have a new cron job that attempts to populate this field based on the `clone_url`. I.e., it will attempt to find associated repos and baked repos based on the clone url and set the `repo_id` field for that column.

brandonroberts:
> Revisiting this and pushed changes that address:
> 
> * Use the `"id = :id", { id }` for sanitizing selecting on id.
> * Makes the commit authors field hidden per @brandonroberts's suggestion.
>   
>   * Does punting on the encryption bits make sense right now? If this field is hidden via our API, that should at least keep people from scraping emails. But we may want to investigate adding encryption at rest for other services that may read this data from the database.
> * Adds a `repo_id` field on the baked repos table that references an id in the repos table.
>   
>   * I'm thinking that we'll have the ETL service have a new cron job that attempts to populate this field based on the `clone_url`. I.e., it will attempt to find associated repos and baked repos based on the clone url and set the `repo_id` field for that column.

Yes, to punting on the encryption. We won't provide any public access to read commit author emails, but more investigation is needed.

I don't think the `repo_id` needs to have a hard reference to the `repos` table. If it does, then we can't insert a baked_repo record for a repo_id that doesn't already exist in our data.

bdougie (approved):
I happy with the decisions made in this PR and agree we can punt on encryption

jpmcb:
> I don't think the repo_id needs to have a hard reference to the repos table. If it does, then we can't insert a baked_repo record for a repo_id that doesn't already exist in our data.

So are you thinking this is ok for now? Or should this column be modified somehow to not directly link to a row in the repos table?

brandonroberts:
> > I don't think the repo_id needs to have a hard reference to the repos table. If it does, then we can't insert a baked_repo record for a repo_id that doesn't already exist in our data.
> 
> So are you thinking this is ok for now? Or should this column be modified somehow to not directly link to a row in the repos table?

It should be modified to not directly link it, as it will fail on insert because of the reference

jpmcb:
@brandonroberts - made the following change:

```diff
+  -- elastic columns
+
+  -- the repo_id field is used to loosely reference an id from the repos table
+  -- but since git repos may have their commits indexed that have not yet been
+  -- indexed into the repos table, there is not direct correlation.
+  -- Joins can happen across the "clone_url" to find repos that have not yet
+  -- synced to baked_repos.
+  repo_id bigint default null,
+
```

I'm not sure if that's the right approach or if we have precedence for a column that can be null that loosely references the id of another table.

brandonroberts (approved):
🍕

jpmcb:
Great! Looks like we have alignment on the data points and decisions made in this PR.

Next steps before this merges:
- get these data changes into the `pizza` oven - https://github.com/open-sauced/pizza/pull/30
- get that version of the `pizza` oven deployed
- merge and release these changes with that deployment of the pizza oven set in the env var

jpmcb:
Force pushed to rebase on `beta` and address `eslint` issues.

This is ready to merge now 🚢

PR #242, opened 2023-08-05
feat: accept highlight type during highlight creation

diivi:
## Description

Allows accepting a "type" for highlights so we can extend highlights to blogs, issues etc.
When I add an issue from the frontend right now, it gets posted as a pull_request anyway.

- [x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI
- [ ] 📦 Chore (Release)
- [ ] ⏩ Revert

## Added tests?

- [ ] 👍 yes

## Added to documentation?

- [ ] 📜 README.md
- [ ] 📓 docs.opensauced.pizza
- [ ] 🍕 dev.to/opensauced
- [ ] 📕 storybook

### DIFF
--- a/src/user/user-highlights.service.ts
+++ b/src/user/user-highlights.service.ts
@@ -204,6 +204,7 @@ export class UserHighlightsService {
       highlight: highlight.highlight,
       title: highlight.title ?? "",
       shipped_at: highlight.shipped_at ? new Date(highlight.shipped_at) : new Date(),
+      type: highlight.type,
     });
 
     return this.userHighlightRepository.save(newUserHighlight);


src/user/user-highlights.service.ts
```diff
@@ -204,6 +204,7 @@ export class UserHighlightsService {
       highlight: highlight.highlight,
       title: highlight.title ?? "",
       shipped_at: highlight.shipped_at ? new Date(highlight.shipped_at) : new Date(),
+      type: highlight.type,
```
diivi, line 207: this should be backward compatible since we have a default for `type` anyway, right?https://github.com/diivi/api.opensauced.pizza/blob/47521cc26beaae3c466b822935f238f29ae2d37e/src/user/dtos/create-user-highlight.dto.ts#L48

bdougie: We can run a migration make a highlight created pull request type. But my assumption is all highlights are type "pull_request"

diivi: Yeah, currently all highlights are `pull_request`

brandonroberts (approved):

src/user/user-highlights.service.ts
brandonroberts: The default value in the database schema is pull_request, so all current records have that value

PR #248, opened 2023-08-09
feat: add source user info to the user notifications

### LINKED ISSUE #247: Feature: add username to the notifications and remove highlight id in the message
This error may due to the notification meta_id being used here. We have access to the user id, but we cache users based on usernames.

We will probably need an API update to include username. 

https://github.com/open-sauced/insights/pull/1478/files#diff-2524a786a0bcb5deadfbb244b5d5ef18304b923e055b9f64c1dd9fd0c695a592R158

We will also don't need to add the meta_id to the message. Weird seeing `239 BekahHW did a thing`

_Originally posted by @bdougie in https://github.com/open-sauced/insights/issues/1529#issuecomment-1670690257_

takanome-dev:
## Description
See [comment below](https://github.com/open-sauced/api/pull/248#issuecomment-1673280598)

- [x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI
- [ ] 📦 Chore (Release)
- [ ] ⏩ Revert

Fixes #247

api-notif-relation-resp

## Added tests?

- [ ] 👍 yes

## Added to documentation?

- [ ] 📜 README.md
- [ ] 📓 docs.opensauced.pizza
- [ ] 🍕 dev.to/opensauced
- [ ] 📕 storybook

### DIFF
--- a/src/user/entities/user-notification.entity.ts
+++ b/src/user/entities/user-notification.entity.ts
@@ -1,8 +1,10 @@
-import { Entity, Column, BaseEntity, PrimaryColumn, CreateDateColumn } from "typeorm";
+import { Entity, Column, BaseEntity, PrimaryColumn, CreateDateColumn, JoinColumn, ManyToOne } from "typeorm";
 import {
   ApiModelProperty,
   ApiModelPropertyOptional,
 } from "@nestjs/swagger/dist/decorators/api-model-property.decorator";
+import { ApiHideProperty } from "@nestjs/swagger";
+import { DbUser } from "../user.entity";
 import { UserNotificationTypes } from "./user-notification.constants";
 
 @Entity({ name: "user_notifications" })
@@ -77,4 +79,12 @@ export class DbUserNotification extends BaseEntity {
     length: 32,
   })
   public meta_id?: string;
+
+  @ApiHideProperty()
+  @ManyToOne(() => DbUser, (user) => user.from_user_notifications)
+  @JoinColumn({
+    name: "from_user_id",
+    referencedColumnName: "id",
+  })
+  public from_user: DbUser;
 }
--- a/src/user/user-notifcation.service.ts
+++ b/src/user/user-notifcation.service.ts
@@ -30,7 +30,8 @@ export class UserNotificationService {
       .innerJoin("users", "users", "user_notifications.user_id=users.id")
       .innerJoinAndSelect("user_notifications.from_user", "from_user")
       .where("user_id = :userId", { userId })
-      .andWhere("user_notifications.type IN (:...userNotificationTypes)", { userNotificationTypes });
+      .andWhere("user_notifications.type IN (:...userNotificationTypes)", { userNotificationTypes })
+      .orderBy("user_notifications.notified_at", "DESC");
 
     queryBuilder.offset(pageOptionsDto.skip).limit(pageOptionsDto.limit);
 
--- a/src/user/user-notifcation.service.ts
+++ b/src/user/user-notifcation.service.ts
@@ -28,6 +28,7 @@ export class UserNotificationService {
 
     queryBuilder
       .innerJoin("users", "users", "user_notifications.user_id=users.id")
+      .innerJoinAndSelect("user_notifications.from_user", "from_user")
       .where("user_id = :userId", { userId })
       .andWhere("user_notifications.type IN (:...userNotificationTypes)", { userNotificationTypes });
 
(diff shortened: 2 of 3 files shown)


takanome-dev:
Since there are already notifications in the db, we will probably need to update them by setting `from_user` to the username.
```sql
UPDATE public.user_notifications
SET from_user = u.login
FROM public.users u
WHERE public.user_notifications.from_user_id = u.id;
```

brandonroberts:
@takanome-dev this is not what want we want to do here. We don't need to duplicate the username in the table. The links I posted in the original issue show how to build a reference to the user information from the users table, so it will get the information and attach it to the notification object.

That way it always reflects the user information based on the ID

takanome-dev:
> The links I posted in the original issue show how to build a reference to the user information from the users table

Oh I didn't get it, thanks for the clarification. So, if I understand better I have to do the following:
- drop the `from_user_id` column in the notifications table
- add `from` column as a foreign key to the `users` table in the notifications table
- add `notifications` column in the `users` table
- join the user with the notifications when retrieving the notifications

Did I get that right, @brandonroberts?

bdougie (approved):

src/user/entities/user-notification.entity.ts
```diff
@@ -77,4 +77,14 @@ export class DbUserNotification extends BaseEntity {
...
   })
   public meta_id?: string;
+
+  @ApiModelProperty({
+    description: "User notification source",
+    example: "bdougie",
+  })
+  @Column({
+    type: "character varying",
+    length: 32,
```
bdougie, line 87: I think this will eventually come back to bite us. The limit on GitHub is 39.

https://www.google.com/search?q=what+is+the+limit+for+github+usernames&oq=what+is+the+limit+for+github+usernames&gs_lcrp=EgZjaHJvbWUyBggAEEUYOTIHCAEQIRigATIHCAIQIRigATIKCAMQIRgWGB0YHjIKCAQQIRgWGB0YHtIBCDY4NDNqMGo3qAIAsAIA&sourceid=chrome&ie=UTF-8

bdougie (changes requested):

brandonroberts:
@takanome-dev not quite 😅 

We're not removing any fields from the table. We're using TypeORM's syntax for building SQL joins with code to associate more information with each user notification we query.

With the notifications, we already have the `from_user_id`, so we want to go fetch the `DbUser` from the `users` table and include it with each notification.

Its a multi-step process to setup.

## Set up the One to Many from User -> User Notification

In the `DbUser` entity, we connect one user to many `DbNotification` records by adding a `OneToMany` relationship

```ts
export class DbUser {
  // other entity fields...

  @ApiHideProperty()
  @OneToMany(() => DbUserNotification, (fromUserNotifications) => fromUserNotifications.from_user)
  public from_user_otifications: DbUserNotification[];
}
```

## Set up the Many to One for UserNotifcation -> User

In the `DbUserNotification` entity, we connect many `DbUserNotification` records back to one user by adding a many to one relationship

```ts
export class DbUserNotifcation {
  // other entity fields...

  @ApiHideProperty()
  @ManyToOne(() => DbUser, (user) => user.from_user_otifications)
  @JoinColumn({
    name: "from_user_id",
    referencedColumnName: "id",
  })
  public from_user: DbUser;
}
```

## Add the field to the query builder

Now,  you can query notifications and request that the `from_user` be populated with the `DbUser` based on the `from_user_id`

```ts
    const queryBuilder = this.baseQueryBuilder(); // user_notifications query builder

    queryBuilder
      .innerJoinAndSelect("user_notifications.from_user", "from_user")
```

TypeORM will attach each notification with a `from_user` property in the returned data.

```
{
  // other user notification fields ...

  from_user_id: 42211
  from_user: {
    login: 'brandonroberts'
    // ... more user fields
  }
}
```

Hopefully this clears things up a bit

takanome-dev:
Wow!! Thanks so much @brandonroberts

src/user/entities/user-notification.entity.ts
takanome-dev: should I update that as well @bdougie?

brandonroberts (approved):

bdougie (approved):
Thanks for the direction on this one @brandonroberts

PR #274, opened 2023-08-23
fix: update BakeRepoDto.wait to boolean

### LINKED ISSUE #273: Bug: Properties with type 'number` should be 'integer'
### Describe the bug

There is a problem with the data types of generated clients, such as [open-sauced/go-api](https://github.com/open-sauced/go-api). They are based on the OpenAPI (Swagger) specification found in this repository.

Currently, most properties that are expected to be numbers, such as IDs, pagination (page, limit), etc. are specified with type `number`, which in Golang translates to `float32`. These properties' types should be updated to `integer`. By doing this change we can use the correct property type without unnecessary casting.

### Steps to reproduce

N/A

### Browsers

_No response_

### Additional context (Is this in dev or production?)

_No response_

### Code of Conduct

### Contributing Docs

- [X] I agree to follow this project's Contribution Docs

cecobask:
## Description

API calls to [POST /v1/bake](https://api.opensauced.pizza/#/Pizza%20oven%20service/Bake%20a%20repository%20with%20the%20pizza%20oven%20microservice) fail with status code 400 because the expected type in the pizza service is boolean.

- [ ] 🍕 Feature
- [x] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI
- [ ] 📦 Chore (Release)
- [ ] ⏩ Revert

Relates to https://github.com/open-sauced/go-api/issues/2

## Steps to reproduce

1. Go to https://api.opensauced.pizza/#/Pizza%20oven%20service/Bake%20a%20repository%20with%20the%20pizza%20oven%20microservice
2. Set environment to production
3. Set the request body to:
```
{
  "url": "https://github.com/open-sauced/insights",
  "wait": "false"
}
```
4. Observe the response status code to be 400
5. Set the request body to:
```
{
  "url": "https://github.com/open-sauced/insights",
  "wait": false
}
```
6. Observe the response status code to be 201

N/A

## Added tests?

- [ ] 👍 yes
- [X] 🙅 no, because they aren't needed

## Added to documentation?

- [ ] 📜 README.md
- [ ] 📓 docs.opensauced.pizza
- [ ] 🍕 dev.to/opensauced
- [ ] 📕 storybook
- [X] 🙅 no documentation needed

## Are there any post-deployment tasks we need to perform?
Update / regenerate the [open-sauced/go-api](https://github.com/open-sauced/go-api) client.

### DIFF
--- a/src/pizza/dtos/baked-repo.dto.ts
+++ b/src/pizza/dtos/baked-repo.dto.ts
@@ -12,7 +12,7 @@ export class BakeRepoDto {
 
   @ApiProperty({
     description: "Option to wait for Pizza service to finish baking repo",
-    type: String,
+    type: Boolean,
     example: false,
   })
   @IsBoolean()
--- a/swagger.yaml
+++ b/swagger.yaml
@@ -6043,7 +6043,7 @@ components:
           description: Repo clone URL
           example: https://github.com/open-sauced/insights
         wait:
-          type: string
+          type: boolean
           description: Option to wait for Pizza service to finish baking repo
           example: false
       required:


brandonroberts:
Thanks @cecobask. The swagger file is generated each time the server starts up, so these changes will be overwritten the next time. Would have to look at changing the Swagger types defined in order to support this.

cecobask:
> Thanks @cecobask. The swagger file is generated each time the server starts up, so these changes will be overwritten the next time. Would have to look at changing the Swagger types defined in order to support this.

Hi @brandonroberts, happy to make this change, but I got a bit lost with defining types as integers in `@ApiProperty`.
I believe TypeScript doesn't have an integer type, but I haven't had much experience with it.
Would that mapping even be possible? 
I do know that [OpenAPI specification](https://swagger.io/docs/specification/data-models/data-types/#numbers) allows for it, not sure how to address this with code generation.

jpmcb:
> are specified with type number, which in Golang translates to float32. These properties' types should be updated to integer.

Hmm that is annoying and yes, we don't want those to be generated as floats since they exist as 

> Would that mapping even be possible?

Good question. Going from our NestJS api -> swagger doc -> OpenAPI generated client -> Go code has proven to be a delicate process and requires abit of magic.

Worse comes to worse, we can always apply a post generation step via a script with something like:
```sh
sed -i "" 's|type: number|type: integer|g' swagger.yaml
```
although that solution will be _very_ brittle since it will blast away any intention of using a "number" as that _should_ have point precision.

I'll take a deeper look at this so we can generate an accurate client.

Thanks for raising this issue ❤️

cecobask:
Thanks, @jpmcb!
I updated my branch to address just the issue with `BakeRepoDto.wait` defined as a string, instead of a boolean.
I've dropped the changes relating to `number` -> `integer`.

> although that solution will be very brittle since it will blast away any intention of using a "number" as that should have point precision.

I agree, this won't scale well if in the future there is a need for properties with float type 😄

jpmcb (changes requested):
Looks like the swagger doc didn't getting updated: we have this update each time the API runs. Can you execute `npm run start` to get the swagger doc updated?

If not, I can add a commit to this PR to address that 😄

cecobask:
> Looks like the swagger doc didn't getting updated: we have this update each time the API runs. Can you execute `npm run start` to get the swagger doc updated?
> 
> If not, I can add a commit to this PR to address that 😄

Done 👍🏻

jpmcb (approved):

brandonroberts (approved):

PR #276, opened 2023-08-24
fix: update full path to swagger file for writing

brandonroberts:
## Description

- Fixes an issue with getting access denied when writing the swagger file when starting up after the deployment

## What type of PR is this? (check all applicable)

- [ ] 🍕 Feature
- [x] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] 🔥 Performance Improvements
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI
- [ ] 📦 Chore (Release)
- [ ] ⏩ Revert

## Added tests?

- [ ] 👍 yes
- [ ] 🙅 no, because they aren't needed
- [ ] 🙋 no, because I need help

## Added to documentation?

- [ ] 📜 README.md
- [ ] 📓 docs.opensauced.pizza
- [ ] 🍕 dev.to/opensauced
- [ ] 📕 storybook
- [ ] 🙅 no documentation needed

PR #279, opened 2023-08-24
chore: Create a generate-swagger.ts script for generating swagger spec

jpmcb:
## Description

This removes generating the swagger spec within the actual app itself and instead offloads it to a `generate-swagger.ts` script that gets invoked via a different nest-js CLI config json file.

- [ ] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [x] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI
- [ ] 📦 Chore (Release)
- [ ] ⏩ Revert

Related to https://github.com/open-sauced/api/pull/276

N/a

## Added tests?

- [ ] 👍 yes

## Added to documentation?

- [ ] 📜 README.md
- [ ] 📓 docs.opensauced.pizza
- [ ] 🍕 dev.to/opensauced
- [ ] 📕 storybook

N/a

### DIFF
--- a/src/generate-swagger.ts
+++ b/src/generate-swagger.ts
@@ -1,0 +1,58 @@
+import path from "node:path";
+import { writeFile } from "node:fs/promises";
+import { NestFactory } from "@nestjs/core";
+import { FastifyAdapter, NestFastifyApplication } from "@nestjs/platform-fastify";
+import { SwaggerModule, DocumentBuilder } from "@nestjs/swagger";
+import { VersioningType } from "@nestjs/common";
+import { Logger } from "nestjs-pino";
+import { major } from "semver";
+import * as yaml from "yaml";
+
+import { AppModule } from "./app.module";
+import { swaggerMarkdownDescription } from "./common/util/swagger";
+
+const generateSwaggerJson = async () => {
+  const app = await NestFactory.create<NestFastifyApplication>(AppModule, new FastifyAdapter({ logger: false }), {
+    bufferLogs: true,
+    rawBody: true,
+  });
+
+  const apiDomain = "api.opensauced.pizza";
+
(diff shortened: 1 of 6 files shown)


src/generate-swagger.ts
```diff
@@ -0,0 +1,58 @@
...
+    .addBearerAuth();
+
+  const document = SwaggerModule.createDocument(app, options.build(), {
+    operationIdFactory: (controllerKey: string, methodKey: string) => methodKey,
+  });
+
+  const yamlSwaggerDoc = yaml.stringify(document);
+
+  // write the yaml swagger doc to the root project directory anytime the server starts
+  await writeFile(path.resolve(process.cwd(), "swagger.yaml"), yamlSwaggerDoc, "utf8");
```
brandonroberts, line 53: Writing the swagger file at the root throws an error on DigitalOcean during deployment. Can we wrap this in an environment variable or what would you recommend?

jpmcb: > Writing the swagger file at the root throws an error on DigitalOcean during deployment.

This change abstracts that completely: by providing a different, non-default Nest JS cli config that is only invoked through `npm run generate:swagger`, we can avoid having to write the file at all when the real server starts. This probably should have been the original change since adding new I/O file behavior to the server on startup was probably not wise on my part.

So now, in a real deployment scenario, we are back to the original behavior without the file being written to the project root but a maintainer or contributor making changes can manually re-generate the swagger doc as a development task when contributing.

brandonroberts (approved):

bdougie (approved):

PR #291, opened 2023-09-04
feat: add user search endpoint to API

OgDev-01:
## Description
This PR adds a new file `filtered-users.dto.ts` in the `src/user/dtos` directory. This file defines a class `FilteredUsersDto` with properties `page`, `username`, and `limit` along with their respective validations and default values. It also includes a `skip` getter method. 

A new file `filtered-users.entity.ts` is added in the `src/user/entities` directory. This file defines a class `DbFilteredUser` with properties `login` and `full_name`.

In the `user.service.ts` file, a new method `findUsersByFilter` is added to the `UserService` class. This method takes a `FilteredUsersDto` object as a parameter and returns a `PageDto` of `DbFilteredUser` entities based on the filter criteria.

In the `user.controller.ts` file, a new route `/search` is added with the `getUserByFilter` method. This route allows searching users based on the filter criteria specified in the `FilteredUsersDto` object.

This PR enhances the functionality of the user service by adding the ability to search users based on specific criteria.

_Generated using [OpenSauced](https://opensauced.ai/)._

- [x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI
- [ ] 📦 Chore (Release)
- [ ] ⏩ Revert

Closes https://github.com/open-sauced/engineering/issues/131

### Example response
Screenshot 2023-09-04 at 16 16 35

### Endpoint 
Screenshot 2023-09-04 at 16 17 51

## Added tests?

- [ ] 👍 yes

## Added to documentation?

- [ ] 📜 README.md
- [ ] 📓 docs.opensauced.pizza
- [ ] 🍕 dev.to/opensauced
- [ ] 📕 storybook

### DIFF
--- a/src/user/dtos/filtered-users.dto.ts
+++ b/src/user/dtos/filtered-users.dto.ts
@@ -1,0 +1,41 @@
+import { ApiPropertyOptional } from "@nestjs/swagger";
+import { Type } from "class-transformer";
+import { IsInt, Min, IsOptional, Max } from "class-validator";
+
+export class FilteredUsersDto {
+  @ApiPropertyOptional({
+    minimum: 1,
+    default: 1,
+    type: "integer",
+  })
+  @Type(() => Number)
+  @IsInt()
+  @Min(1)
+  @IsOptional()
+  readonly page?: number = 1;
+
+  @ApiPropertyOptional({
+    description: "Username search query to filter from the list of users",
+    type: "string",
+  })
+  @IsOptional()
+  @Type(() => String)
+  readonly username?: string = "";
+
+  @ApiPropertyOptional({
+    minimum: 1,
+    maximum: 1000,
+    default: 10,
+    type: "integer",
+  })
+  @Type(() => Number)
+  @IsInt()
+  @Min(1)
+  @Max(10000)
+  @IsOptional()
+  readonly limit?: number = 50;
+
+  get skip(): number {
+    return ((this.page ?? 1) - 1) * (this.limit ?? 50);
+  }
+}
--- a/src/user/entities/filtered-users.entity.ts
+++ b/src/user/entities/filtered-users.entity.ts
@@ -1,0 +1,15 @@
+import { ApiModelProperty } from "@nestjs/swagger/dist/decorators/api-model-property.decorator";
+
+export class DbFilteredUser {
+  @ApiModelProperty({
+    description: " User Login",
+    example: "bdougie",
+  })
+  public login?: string;
+
+  @ApiModelProperty({
+    description: "Users fullname",
+    example: "Brian Douglas",
+  })
+  public full_name: string;
+}
--- a/src/user/services/user.service.ts
+++ b/src/user/services/user.service.ts
@@ -14,6 +14,8 @@ import { DbTopUser } from "../entities/top-users.entity";
 import { TopUsersDto } from "../dtos/top-users.dto";
 import { PageDto } from "../../common/dtos/page.dto";
 import { PageMetaDto } from "../../common/dtos/page-meta.dto";
+import { DbFilteredUser } from "../entities/filtered-users.entity";
+import { FilteredUsersDto } from "../dtos/filtered-users.dto";
 
 @Injectable()
 export class UserService {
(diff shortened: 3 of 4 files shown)


OgDev-01:
In the future, we can add more query by changing the username params to something similar to githubs query.

### Example 
`username` would be `q` in this instance
This way, we can add several queries like `username: bdougie` or `full_name: brian douglas`

src/user/dtos/filtered-users.dto.ts
```diff
@@ -0,0 +1,41 @@
...
+  @IsOptional()
+  readonly page?: number = 1;
+
+  @ApiPropertyOptional({
+    description: "Username search query to filter from the list of users",
+    type: "string",
+  })
+  @IsOptional()
+  @Type(() => String)
+  readonly username?: string = "";
```
brandonroberts, line 23: This should be required and a minimum of 3 characters to prevent overloading the API with wildcard searches.

src/user/dtos/filtered-users.dto.ts
```diff
@@ -0,0 +1,41 @@
...
+  @ApiPropertyOptional({
+    minimum: 1,
+    maximum: 1000,
+    default: 10,
+    type: "integer",
+  })
+  @Type(() => Number)
+  @IsInt()
+  @Min(1)
+  @Max(10000)
```
brandonroberts, line 34:
```suggestion
  @Max(100)
```

src/user/entities/filtered-users.entity.ts
```diff
@@ -0,0 +1,15 @@
+import { ApiModelProperty } from "@nestjs/swagger/dist/decorators/api-model-property.decorator";
+
+export class DbFilteredUser {
+  @ApiModelProperty({
+    description: " User Login",
```
brandonroberts, line 5:
```suggestion
    description: "User Login",
```

src/user/services/user.service.ts
```diff
@@ -177,6 +179,31 @@ export class UserService {
     return item;
   }
 
+  async findUsersByFilter(pageOptionsDto: FilteredUsersDto): Promise<PageDto<DbFilteredUser>> {
+    const queryBuilder = this.baseQueryBuilder();
+
+    const { username, limit } = pageOptionsDto;
+
+    if (!username) {
+      throw new NotFoundException();
```
brandonroberts, line 188:
```suggestion
      throw new BadRequestException();
```

src/user/services/user.service.ts
```diff
@@ -177,6 +179,31 @@ export class UserService {
...
+    }
+
+    if (username) {
```
brandonroberts, line 191: The `if` check isn't needed if we're throwing an error above

```suggestion

```

src/user/dtos/filtered-users.dto.ts
```diff
@@ -0,0 +1,41 @@
...
+  @Type(() => Number)
+  @IsInt()
+  @Min(1)
+  @IsOptional()
+  readonly page?: number = 1;
+
+  @ApiPropertyOptional({
+    description: "Username search query to filter from the list of users",
+    type: "string",
+  })
```
jpmcb, line 20: Can you run `npm run generate:swagger` to add this to the swagger doc?

src/user/dtos/filtered-users.dto.ts
jpmcb: I'd imagine we don't want to ever send an empty string `""` since that would always return all users. +1 to what Brandon said: we should have a minimum bar for what the API can accept.

src/user/dtos/filtered-users.dto.ts
```diff
@@ -0,0 +1,41 @@
...
+    maximum: 1000,
+    default: 10,
+    type: "integer",
+  })
+  @Type(() => Number)
+  @IsInt()
+  @Min(1)
+  @Max(100)
+  @IsOptional()
+  readonly limit?: number = 50;
```
jpmcb, line 36: 50 is alot to return on default: my gut is just a few as default (like 5) but maybe that's left open for the frontend to pair down

src/user/services/user.service.ts
```diff
@@ -177,6 +179,31 @@ export class UserService {
...
+      queryBuilder
+        .select(["users.login as login", "users.name as full_name"])
+        .where(`LOWER(users.login) LIKE :username`)
```
jpmcb, line 194: Hmmm. Interesting. Are username's case sensitive? My guess is "no" since github username's are not case sensitive. We'll just want to make sure so we don't accidentally squash duplicate users who have case sensitive usernames. I.e., returning `[brian, brian]` for users `Brian` and `bRiAn`

OgDev-01:
> I'd imagine we don't want to ever send an empty string "" since that would always return all users

We have a check in place for this case which throws a new `BadRequestException()` error if the username is an empty string

src/user/dtos/filtered-users.dto.ts
OgDev-01: Ok 👍

src/user/services/user.service.ts
jpmcb: Sounds great!! I'm more just musing here 😂 if we're returning the case sensitive form from the database, that works!!! 👏🏼

jpmcb (approved):
👏🏼

PR #294, opened 2023-09-05
feat: Bake repos when new repos added to insights or fetched

jpmcb:
## Description

This proof of concept is twofold:

- When an insights paged is fetched, it calls the pizza oven service for those repos, getting new commit data.
- When an insights page is created, it calls the pizza oven service for those repos, getting new commit data or backfill all author/repo commit data (since it may be a new repo we have not gotten commit data for)

This essentially flips the paradigm to only get new commit data when there is an active user request for it.

Curious to hear other's thoughts on this approach since it would be abit different then how we've backfilled data in the past.

- [x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI
- [ ] 📦 Chore (Release)
- [ ] ⏩ Revert

Related to:
- https://github.com/open-sauced/etl.opensauced.pizza/issues/83
- https://github.com/open-sauced/etl.opensauced.pizza/issues/79

N/a

## Added tests?

- [ ] 👍 yes

## Added to documentation?

- [ ] 📜 README.md
- [ ] 📓 docs.opensauced.pizza
- [ ] 🍕 dev.to/opensauced
- [ ] 📕 storybook

### DIFF
--- a/src/insight/insight-repo.service.ts
+++ b/src/insight/insight-repo.service.ts
@@ -3,16 +3,27 @@ import { Repository } from "typeorm";
 import { InjectRepository } from "@nestjs/typeorm";
 
 import { RepoInfo } from "../repo/dtos/repo-info.dto";
+import { PizzaOvenService } from "../pizza/pizza-oven.service";
+import { BakeRepoDto } from "../pizza/dtos/baked-repo.dto";
 import { DbInsightRepo } from "./entities/insight-repo.entity";
 
 @Injectable()
 export class InsightRepoService {
   constructor(
     @InjectRepository(DbInsightRepo, "ApiConnection")
-    private insightRepoRepository: Repository<DbInsightRepo>
+    private insightRepoRepository: Repository<DbInsightRepo>,
+    private pizzaOvenService: PizzaOvenService
   ) {}
 
   async addInsightRepo(insightId: number, repo: RepoInfo) {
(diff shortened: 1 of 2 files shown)


bdougie:
Currently in the app, we are not showing any commit data and this would enable us to provide deeper insights for users who request it. 

Basically those with highlights and insights have accounts on opensauced and creating signals for us to know what projects matter the most. I think the paradigm makes sense, as long if there is still a backdoor to get repos baked via the CLI on demand. That interaction can eventually be gated via log in, but I think this is a step in the right direction.

bdougie (approved):
I think this makes a lot of sense. Fetching new data for repos that are viewed the most. This pattern could be mirrored in our filtered topic "explore" endpoints as well, but perhaps that is better to do on a cron, rather than on page view fetching.

jpmcb:
> reading again and seeing this is enabling insights and not highlights in this PR.

This could pretty easily be added to highlights in the PR as well. If we think that's a good signal as well, we can go for it!

brandonroberts:
I think it's a good step also. We also talked about adding a `last_viewed` on an insight page record also to give us some data on which pages/repos are actively being watched.

brandonroberts (approved):

PR #293, opened 2023-09-05
feat: add migrations and initial endpoints for user lists

brandonroberts:
## Description

- Adds database migrations for user lists
- Adds CRUD endpoints for managing user lists

## What type of PR is this? (check all applicable)

- [x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] 🔥 Performance Improvements
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI
- [ ] 📦 Chore (Release)
- [ ] ⏩ Revert

## Related Tickets & Documents

Related to: https://github.com/open-sauced/engineering/issues/107

## Mobile & Desktop Screenshots/Recordings

image

Sample output

```json
{
  "id": "uuid-v4",
  "user_id": 237133,
  "name": "JavaScript Developers",
  "is_public": false,
  "created_at": "2022-10-19 13:24:51.000000",
  "updated_at": "2022-10-19 13:24:51.000000",
  "deleted_at": "2022-10-19 13:24:51.000000"
}
```

## Added tests?

- [ ] 👍 yes
- [ ] 🙅 no, because they aren't needed
- [ ] 🙋 no, because I need help

## Added to documentation?

- [ ] 📜 README.md
- [ ] 📓 docs.opensauced.pizza
- [ ] 🍕 dev.to/opensauced
- [ ] 📕 storybook
- [ ] 🙅 no documentation needed

PR #297, opened 2023-09-06
feat: User-lists collaborators CRUD endpoints

jpmcb:
## Description

- Creates the following crud routes:
  - `/v1/lists/{id}/contributors` - [GET] Get paginated set of contributors in the list
  - `/v1/lists/{id}/contributors` - [POST] Add new contributor to an existing list
  - `/v1/lists/{id}/contributors` - [DELETE] Remove contributor from an existing list
  - `/v1/lists/contributors` - [GET] Get paginated set of contributors to display when creating a list (w/filters)
- Sets `"integer"` type for swagger doc creation for user-list entities (this was a doubley nested array in the swagger doc before this change):

Screenshot 2023-09-06 at 4 12 13 PM

- [x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI
- [ ] 📦 Chore (Release)
- [ ] ⏩ Revert

Related to https://github.com/open-sauced/api/pull/293

## Added tests?

- [ ] 👍 yes

## Added to documentation?

- [] 📜 README.md
- [ ] 📓 docs.opensauced.pizza
- [ ] 🍕 dev.to/opensauced
- [ ] 📕 storybook

### DIFF
--- a/src/user-lists/entities/user-list-contributor.entity.ts
+++ b/src/user-lists/entities/user-list-contributor.entity.ts
@@ -5,6 +5,7 @@ import {
   Entity,
   JoinColumn,
   ManyToOne,
+  PrimaryColumn,
   PrimaryGeneratedColumn,
 } from "typeorm";
 import { ApiHideProperty } from "@nestjs/swagger";
@@ -15,15 +16,15 @@ import {
 
 import { DbUser } from "../../user/user.entity";
 
-@Entity({ name: "user_organizations" })
+@Entity({ name: "user_list_contributors" })
 export class DbUserListContributor {
   @ApiModelProperty({
-    description: "User organization identifier",
-    example: 196,
-    type: "integer",
+    description: "User list contributor identifier",
+    example: "uuid-v4",
   })
+  @PrimaryColumn()
   @PrimaryGeneratedColumn()
-  public id!: number;
+  public id!: string;
 
   @ApiModelProperty({
     description: "User identifier",
@@ -36,7 +37,6 @@ export class DbUserListContributor {
   @ApiModelProperty({
     description: "List identifier",
     example: "uuid-v4",
-    type: "integer",
   })
   @Column()
   public list_id!: string;
@@ -65,4 +65,15 @@ export class DbUserListContributor {
     referencedColumnName: "id",
   })
   public user_list_contributor!: DbUser;
+
+  @ApiModelProperty({
+    description: "User list collaborator's login",
+    example: "bdougie",
+  })
+  @Column({
+    type: "text",
+    select: false,
+    insert: false,
+  })
+  public login?: string;
 }
--- a/src/user-lists/entities/user-list.entity.ts
+++ b/src/user-lists/entities/user-list.entity.ts
@@ -19,7 +19,7 @@ import { ApiHideProperty } from "@nestjs/swagger";
 
 import { DbUser } from "../../user/user.entity";
 
-@Entity({ name: "insights" })
+@Entity({ name: "user_lists" })
 export class DbUserList extends BaseEntity {
   @ApiModelProperty({
     description: "User List identifier",
--- a/src/user-lists/user-list.controller.ts
+++ b/src/user-lists/user-list.controller.ts
@@ -1,4 +1,4 @@
-import { Body, Controller, Delete, Get, Param, ParseIntPipe, Patch, Post, UseGuards } from "@nestjs/common";
+import { Body, Controller, Delete, Get, Param, ParseIntPipe, Patch, Post, Query, UseGuards } from "@nestjs/common";
 import {
   ApiOperation,
   ApiOkResponse,
@@ -10,17 +10,41 @@ import {
   ApiParam,
 } from "@nestjs/swagger";
 
+import { DbUser } from "../user/user.entity";
+import { PageOptionsDto } from "../common/dtos/page-options.dto";
+import { ApiPaginatedResponse } from "../common/decorators/api-paginated-response.decorator";
+import { PageDto } from "../common/dtos/page.dto";
 import { UserId } from "../auth/supabase.user.decorator";
 import { SupabaseGuard } from "../auth/supabase.guard";
 
 import { CreateUserListDto } from "./dtos/create-user-list.dto";
 import { DbUserList } from "./entities/user-list.entity";
 import { UserListService } from "./user-list.service";
+import { DbUserListContributor } from "./entities/user-list-contributor.entity";
+import { CollaboratorsDto } from "./dtos/collaborators.dto";
+import { FilterListContributorsDto } from "./dtos/filter-contributors.dto";
 
 @Controller("lists")
 @ApiTags("User Lists service")
 export class UserListController {
   constructor(private readonly userListService: UserListService) {}
+
+  @Get("/")
+  @ApiOperation({
+    operationId: "getListsForUser",
+    summary: "Gets lists for the authenticated user",
+  })
+  @ApiBearerAuth()
+  @UseGuards(SupabaseGuard)
+  @ApiOkResponse({ type: DbUserList })
+  @ApiNotFoundResponse({ description: "Unable to get user lists" })
+  @ApiBadRequestResponse({ description: "Invalid request" })
+  async getListsForUser(
+    @UserId() userId: number,
+    @Query() pageOptionsDto: PageOptionsDto
+  ): Promise<PageDto<DbUserList>> {
+    return this.userListService.findAllByUserId(pageOptionsDto, userId);
+  }
 
   @Post("/")
   @ApiOperation({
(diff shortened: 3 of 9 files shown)


src/user-lists/user-list.service.ts
```diff
@@ -82,11 +82,38 @@ export class UserListService {
...
   async updateUserList(listId: string, highlight: Partial<DbUserList>) {
     return this.userListRepository.update(listId, highlight);
   }
 
   async deleteUserList(listId: string) {
     return this.userListRepository.softDelete(listId);
   }
+
+  async findAllContributors(pageOptionsDto: PageOptionsDto): Promise<PageDto<DbUserListContributor>> {
+    const queryBuilder = this.userListContributorRepository.createQueryBuilder("user_list_contributors");
```
brandonroberts, line 98: I think this should be a different query to get a list of all available contributors/users. We have users for a specific list covered in the one below. 

Thoughts?

src/user-lists/user-list.service.ts
```diff
@@ -82,11 +82,38 @@ export class UserListService {
     return this.userListContributorRepository.save(newUserListContributor);
   }
 
+  async deleteUserListContributor(userId: number) {
```
brandonroberts, line 85: I think this should be a `string`, as the user list contributor record ID is a uuid

src/user-lists/user-list.service.ts
```diff
@@ -82,11 +82,38 @@ export class UserListService {
...
+    });
+  }
+
+  async findContributorsByListId(
+    pageOptionsDto: PageOptionsDto,
+    listId: string
+  ): Promise<PageDto<DbUserListContributor>> {
+    const queryBuilder = this.userListContributorRepository.createQueryBuilder("user_list_contributors");
+
+    queryBuilder.where("user_list_contributors.list_id = :listId", { listId });
```
brandonroberts, line 112: Same as above with joining the `users` table

src/user-lists/user-list.controller.ts
```diff
@@ -102,4 +107,74 @@ export class UserListController {
...
+    @Param("id") id: string
+  ): Promise<DbUserListContributor[]> {
+    const contributors = updateCollaboratorsDto.contributors.map(async (contributorId) =>
+      this.userListService.addUserListContributor(id, contributorId)
+    );
+
+    return Promise.all(contributors);
+  }
+
+  @Delete("/:id/contributors")
```
brandonroberts, line 166: We should be more specific here to only delete the individual record. The `:id` is the list ID

```suggestion
  @Delete("/:id/contributors/:userListContributorId")
```

brandonroberts (changes requested):

src/user-lists/user-list.service.ts
jpmcb: Ah: so this really should be querying for all `users.id` since a `user_list_contributors.user_id` is the relationship to be made? The way it's written now, this will paginate all of the rows in `user_list_contributors`

src/user-lists/user-list.service.ts
jpmcb: Whoops, this should really be:

```ts
async deleteUserListContributor(id: number)
```

since the intent is the delete the contributor by its id

src/user-lists/user-list.service.ts
jpmcb: So we're not really returning "list contributors", but rather, we're finding the real `users` and getting those? Sounds like `user_list_contributors` is more of a "bridge" table in order to get `users`?

src/user-lists/entities/user-list-contributor.entity.ts
```diff
@@ -15,15 +16,15 @@ import {
...
-    description: "User organization identifier",
-    example: 196,
-    type: "integer",
+    description: "User list contributor identifier",
+    example: "uuid-v4",
   })
+  @PrimaryColumn()
   @PrimaryGeneratedColumn()
-  public id!: number;
+  public id!: string;
```
jpmcb, line 27: Since the db entry for this is:
https://github.com/open-sauced/api/blob/2d08206998303ce31b4990b9ca91dd9239a8734b/migrations/20230905145301_add_user_list_contributors.sql#L4
with a uuid string, changed it to reflect in the type ORM as well

src/user-lists/entities/user-list-contributor.entity.ts
```diff
@@ -15,15 +16,15 @@ import {
 
 import { DbUser } from "../../user/user.entity";
 
-@Entity({ name: "user_organizations" })
+@Entity({ name: "user_list_contributors" })
```
jpmcb, line 19: I'm assuming this was a copy paste error? I was having weird behavior where these columns were getting put into the `user_organizations` table. I'm assuming since the table name is:

https://github.com/open-sauced/api/blob/2d08206998303ce31b4990b9ca91dd9239a8734b/migrations/20230905145301_add_user_list_contributors.sql#L1

that it should also be reflected here

src/user-lists/entities/user-list.entity.ts
```diff
@@ -19,7 +19,7 @@ import { ApiHideProperty } from "@nestjs/swagger";
 
 import { DbUser } from "../../user/user.entity";
 
-@Entity({ name: "insights" })
+@Entity({ name: "user_lists" })
```
jpmcb, line 22: Same here: seemed that the table name "insights" wasn't right

https://github.com/open-sauced/api/blob/2d08206998303ce31b4990b9ca91dd9239a8734b/migrations/20230905145101_add_user_lists.sql#L1

src/user-lists/user-list.controller.ts
```diff
@@ -10,18 +10,40 @@ import {
...
 @Controller("lists")
 @ApiTags("User Lists service")
 export class UserListController {
   constructor(private readonly userListService: UserListService) {}
 
+  @Get("/")
+  @ApiOperation({
+    operationId: "getListsForUser",
+    summary: "Gets lists for the authenticated user",
+  })
```
jpmcb, line 34: Using this:

```sh
curl -X 'GET' \
  'http://localhost:3001/v1/lists?page=1&limit=10&range=30' \
  -H 'Authorization: Bearer <redacted>'
```

returns the two lists I made during playing around with this in the beta env
```json
{
  "data": [
    {
      "id": "17e0f1c2-c556-4734-bfc4-86d43efc4087",
      "name": "My real list",
      "is_public": false,
      "created_at": "2023-09-07T22:42:30.680Z",
      "updated_at": "2023-09-07T22:42:30.680Z",
      "deleted_at": null
    },
    {
      "id": "43fdd609-4936-4ff6-8083-199f46871262",
      "name": "John's tmp List",
      "is_public": false,
      "created_at": "2023-09-07T22:30:10.167Z",
      "updated_at": "2023-09-07T22:30:10.167Z",
      "deleted_at": null
    }
  ],
  "meta": {
    "page": 1,
    "limit": 10,
    "itemCount": 2,
    "pageCount": 1,
    "hasPreviousPage": false,
    "hasNextPage": false
  }
}
```

src/user-lists/user-list.service.ts
```diff
@@ -82,11 +83,51 @@ export class UserListService {
...
+    });
+  }
+
+  async findContributorsByListId(pageOptionsDto: PageOptionsDto, listId: string) {
+    const queryBuilder = this.userListContributorRepository.createQueryBuilder("user_list_contributors");
+
+    queryBuilder
+      .innerJoin("users", "users", "user_list_contributors.user_id=users.id")
+      .addSelect("users.login", "user_lists_login")
+      .where("user_list_contributors.list_id = :listId", { listId });
```
jpmcb, line 124: @brandonroberts I'm confused how to get the `user_lists_login` to appear in the results:

```sh
curl -X 'GET' \
  'http://localhost:3001/v1/lists/17e0f1c2-c556-4734-bfc4-86d43efc4087/contributors?page=1&limit=10&range=30' \
  -H 'Authorization: Bearer <redacted>'
```

gets me

```json
{
  "data": [
    {
      "id": "72211ba0-8036-446d-98af-f5436ba11f4f",
      "user_id": 42211,
      "list_id": "17e0f1c2-c556-4734-bfc4-86d43efc4087",
      "created_at": "2023-09-07T22:42:30.793Z"
    },
    {
      "id": "c6f317ca-fadb-4ca6-b3a7-1a5686cc4ec5",
      "user_id": 23109390,
      "list_id": "17e0f1c2-c556-4734-bfc4-86d43efc4087",
      "created_at": "2023-09-07T22:42:31.002Z"
    }
  ],
  "meta": {
    "page": 1,
    "limit": 10,
    "itemCount": 2,
    "pageCount": 1,
    "hasPreviousPage": false,
    "hasNextPage": false
  }
}
```

which are the two users I added. But it's missing the user's login. I'm guessing it's because this route expects the `DbUserListContributors` back and is squashing extra data not apart of that type. Any ideas or other places we tack on data?

src/user-lists/user-list.service.ts
jpmcb: I think I'm still abit confused what this route is supposed to do. In my eyes, `/v1/lists/contributors [GET]` would return all contributors across all lists (what this route does right now).

But if we want all _possible_ contributors (i.e., all current users), wouldn't we just use the `/v1/users/search` endpoint?

jpmcb (commented):
@brandonroberts - this is ready for another review 💪🏼 let me know if there's anything that's not making sense! Tested this out on the dev database and is looking good!

src/user-lists/entities/user-list-contributor.entity.ts
brandonroberts: 🤦  Yep!

src/user-lists/user-list.service.ts
brandonroberts: We could use `/v1/users/search` but I'm not sure we would need the same filters there as for a contributor search. I don't think we need an endpoint to return all contributors across lists because each list is tied to the authenticated user.

src/user-lists/user-list.service.ts
brandonroberts: In order to see the derived field, it must be on the entity also. Here is an example of a field that's on the entity but. not in the database.

https://github.com/open-sauced/api/blob/beta/src/user/entities/user-highlight.entity.ts#L125-L134

In this case the entity field would be `login` and the alias would be `user_list_contributors_login` because TypeORM uses the entity table name as a prefix on each property in the entity.

jpmcb:
A few changes with the last force push:

- Added `FilterListContributorsDto` which is used as to filter users for the `v1/lists/contributors` get endpoint
  - This endpoint now accepts the Dto filter for "location", "timezone", "pr_velocity". The PR velocity query uses the last 30 days of PRs for a given user and rejects users who have a PR velocity of 0 (since this indicates they have no PRs). Feedback on this bit might be useful. In my testing I found this query to be _abit_ gnarly since it's perform a pretty broad select on potentially a large number of users (depending on how the other filters are configured). I wonder if a better way would be for us to have a field in the database for the `users` table that denotes PR velocity instead of trying to calculate it at runtime on each query.
  - Did quite abit of testing and it all seems to work well. You'll also notice I dropped the `INTEGER` bit of the PR velocity query since this would drop decimal precision for users who have really fast PR velocity (like Brandon's is in the `0.6` range) and ignore them with the `!= 0` clause.
- Adds the derived property for the `user_list_contributors_login`

brandonroberts:
@jpmcb I think having a regularly updated field like `recent_pr_velocity` would be good also. If we're going to have multiple time ranges, we'll have to keep it dynamic and make sure the DB indexes are keeping queries efficient. 

I think its a good step for these endpoints!

brandonroberts (approved):

OgDev-01 (approved):
LGTM

PR #340, opened 2023-10-02
feat: add users timezone endpoint

### LINKED ISSUE #337: feat: add endpoint to get timezones based on data from our database
Curious if we should only show times zones that returns results. Is that possible?

_Originally posted by @bdougie in https://github.com/open-sauced/app/issues/1747#issuecomment-1739822386_

OgDev-01:
## Description
This PR:
- adds endpoint to fetch timezones from the users table

- [x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI
- [ ] 📦 Chore (Release)
- [ ] ⏩ Revert

closes #337 
related to https://github.com/open-sauced/app/pull/1747

image

## Added tests?

- [ ] 👍 yes

## Added to documentation?

- [ ] 📜 README.md
- [ ] 📓 docs.opensauced.pizza
- [ ] 🍕 dev.to/opensauced
- [ ] 📕 storybook


package.json
```diff
@@ -65,6 +65,7 @@
     "@nestjs/terminus": "^9.2.2",
     "@nestjs/typeorm": "^9.0.1",
     "@supabase/supabase-js": "^1.35.6",
+    "axios": "^1.5.1",
```
brandonroberts, line 68: Why are we bringing in axios?

src/user/user.controller.ts
```diff
@@ -107,4 +108,14 @@ export class UserController {
   async getUsersByFilter(@Query() pageOptionsDto: FilteredUsersDto): Promise<PageDto<DbFilteredUser>> {
     return this.userService.findUsersByFilter(pageOptionsDto);
   }
+
+  @Get("/timezones")
```
brandonroberts, line 112: We should put this in the user-list.controller, so its `/lists/timezones` instead of `/user/timezones`

src/user/services/user.service.ts
```diff
@@ -326,4 +327,18 @@ export class UserService {
 
     return item;
   }
+
+  async getAllTimezones(): Promise<DbTimezone[]> {
```
brandonroberts, line 331: Let's move this to the user-list.service

brandonroberts (changes requested):

package.json
OgDev-01: I was getting a package not found error... but i think it is resolved now

brandonroberts:
@OgDev-01 remove the npm-shrinkwrap changes also

jpmcb (commented):
I think code wise this looks great and don't let me block on this - but I'm curious however where we get the "source of truth" for our user's time-zones?

I worry about drift and nasty time bugs: I.e., what's stopping me from entering "Mountain Time" (which is a common way of referring to "Mountain Standard Time") when there's the possibility that what I really am looking for is "Mountain Daylight Time" which is used in the United States for daylights-savings switches in the winter.

Since, within our API, the timezone is just a string:

https://github.com/open-sauced/api/blob/bbb197c1f31697e67a802ef23d87a9df968b9446/src/user/user.entity.ts#L425-L433

and in our database table, it's just a varchar:

https://github.com/open-sauced/api/blob/bbb197c1f31697e67a802ef23d87a9df968b9446/migrations/20220510165555_add_users.sql#L30

getting all timezones is really just getting all free-text fields from the 'users' table. 

This rambling is probably a separate issue and something we need to have consensus on at some point in order to prevent timezone based bugs. Ideally, the API would introspect some source of truth for timezones and service that to the frontend.

jpmcb:
Digging in abit more on my above comment ^ looks like the frontend uses our `lib/util/timezones` in the user settings:

https://github.com/open-sauced/app/blob/beta/components/organisms/UserSettingsPage/user-settings-page.tsx#L306C1-L310C22

which pulls from a constant time zone set we have:

https://github.com/open-sauced/app/blob/504d7e2499a6b2a9dd5beaa560dd9ca7aa6af4c8/lib/utils/timezones.ts

In the future, i think the API should do something similar (or better yet, have a shared opensauced-sdk that _both_ the API and the frontend can use to have the same source of truth)

jpmcb:
Jinx @brandonroberts !! Thanks for the info: again, not blocking, just making comment and hoping to avoid time bugs in the future.

jpmcb (approved):

src/user-lists/user-list.service.ts
```diff
@@ -187,4 +194,18 @@ export class UserListService {
...
   }
+
+  async getAllTimezones(): Promise<DbTimezone[]> {
+    const queryBuilder = this.baseUserQueryBuilder();
+
+    queryBuilder
+      .select("users.timezone as timezone")
+      .where("users.timezone IS NOT NULL")
+      .andWhere("users.timezone != ''")
+      .groupBy("users.timezone");
```
nickytonline, line 205: Refactor (non-blocking): It results in the same thing in this scenario, but would `DISTINCT` not make more sense?

nickytonline (approved):
:shipit:

src/user-lists/user-list.service.ts
OgDev-01: done ✅

nickytonline (approved):
Thanks Sunday! :shipit:

brandonroberts (approved):

PR #346, opened 2023-10-05
fix: Refactors `::DATE` casts to `::TIMESTAMP` casting

jpmcb:
## Description

Throughout our psql calls, we had been using `::DATE` casts to filter queries based on sliding time windows.

These would truncate down ISO timestamps to remove the "time" precision from the date/time strings, leaving essentially just the date. However, when compared to _other_ timestamps we carry data for, we essentially re-cast the date to a timestamp with empty "time".

For example, when getting pull requests for a contributor:

https://github.com/open-sauced/api/blob/84a0e3c01be9d1bef36e219a9b66f265419478d0/src/pull-requests/pull-request.service.ts#L83

given the "startDate" is "NOW()", this would truncate the date ISO string down to something like `2023-10-05` DATE. This is equivalent to the timestamp `2023-10-05 00:00:00Z` which, when compared to _other_ timestamps within the current day, like for example, `2023-10-05 12:30:00`, it would get dropped even though it is well within the sliding time window.

This needs some pretty thorough end to end testing but upon some arbitrary tests, this fixes some issues we saw with recent PRs and things not showing up for people within their profiles for the last day.

- [ ] 🍕 Feature
- [x] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI
- [ ] 📦 Chore (Release)
- [ ] ⏩ Revert

Fixes https://github.com/open-sauced/app/issues/1800

## Added tests?

- [ ] 👍 yes

## Added to documentation?

- [ ] 📜 README.md
- [ ] 📓 docs.opensauced.pizza
- [ ] 🍕 dev.to/opensauced
- [ ] 📕 storybook

Again, need to validate and verify everything is still ok, especially when this lands on beta

### DIFF
--- a/src/common/filters/repo-filter.service.ts
+++ b/src/common/filters/repo-filter.service.ts
@@ -51,8 +51,8 @@ export class RepoFilterService {
           FROM "pull_requests" "spam_pull_requests"
           WHERE
             'spam' = ANY("spam_pull_requests"."label_names")
-            AND '${startDate}'::DATE >= "spam_pull_requests"."updated_at"
-            AND '${startDate}'::DATE - INTERVAL '${range} days' <= "spam_pull_requests"."updated_at"
+            AND '${startDate}'::TIMESTAMP >= "spam_pull_requests"."updated_at"
+            AND '${startDate}'::TIMESTAMP - INTERVAL '${range} days' <= "spam_pull_requests"."updated_at"
         )
       `,
         {},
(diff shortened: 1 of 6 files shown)


brandonroberts:
I did some local testing and it works for the most part. I think we should keep the `DATE` wherever we're using `COALESCE` with some dates. It throws an error using a `TIMESTAMP` on those

```
"COALESCE types interval and integer cannot be matched"
```

jpmcb:
Removed the `TIMESTAMP` casts from where we calculate velocities (since those intervals didn't work with the integer types)

brandonroberts (approved):

bdougie (approved):
LGTM

PR #361, opened 2023-10-10
feat: add flag for featured insights, insights count to user session

brandonroberts:
## Description

- Fetches featured insight pages for users who have not created any insight pages
- Adds insight count total to user session data
- Adds migration to insights table to setting featured insight pages

- [x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI
- [ ] 📦 Chore (Release)
- [ ] ⏩ Revert

Related to: https://github.com/open-sauced/app/issues/1673

## Added tests?

- [ ] 👍 yes

## Added to documentation?

- [ ] 📜 README.md
- [ ] 📓 docs.opensauced.pizza
- [ ] 🍕 dev.to/opensauced
- [ ] 📕 storybook

### DIFF
--- a/src/insight/insights.service.ts
+++ b/src/insight/insights.service.ts
@@ -69,6 +69,15 @@ export class InsightsService {
     queryBuilder
       .where("insights.user_id = :userId", { userId })
       .orWhere(
+        `(
+          (((SELECT COUNT(id) FROM insight_repos where user_id=:userId)=0)
+          AND
+          (is_featured=true AND insights.deleted_at IS NULL))
+        )
+      `,
+        { userId }
+      )
+      .orWhere(
         `:userId IN (
           SELECT user_id
           FROM insight_members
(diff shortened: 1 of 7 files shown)


src/insight/insights.service.ts
```diff
@@ -68,6 +68,15 @@ export class InsightsService {
 
     queryBuilder
       .where("insights.user_id = :userId", { userId })
+      .orWhere(
+        `(
+          (((SELECT COUNT(id) FROM insight_repos where user_id=:userId)=0)
+          AND
+          (is_featured=true AND insights.deleted_at IS NULL))
+        )
```
jpmcb, line 76: Just trying to understand this: so this will return insight pages that are featured for a user that have not been deleted ... ?

jpmcb (approved):

src/insight/insights.service.ts
brandonroberts: Yes, but only if they haven't created any insight pages, which is typically new users. This way we can push featured insight pages into their list and remove them on our side with the `is_featured` flag

jpmcb: Ahhhhh I see i see, I was missing the part where we would push them insight pages. Makes sense 👍🏼 :shipit:

PR #399, opened 2023-10-30
feat: add tier.run service and check if user is registered

brandonroberts:
## Description

- Adds service to check/register users through tier.run/Stripe for feature flag support
- Adds new environment variables example config
- Updates test to mock TierService

- [x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI
- [ ] 📦 Chore (Release)
- [ ] ⏩ Revert

## Added tests?

- [ ] 👍 yes

## Added to documentation?

- [ ] 📜 README.md
- [ ] 📓 docs.opensauced.pizza
- [ ] 🍕 dev.to/opensauced
- [ ] 📕 storybook

### DIFF
--- a/src/tier/tier.service.ts
+++ b/src/tier/tier.service.ts
@@ -8,7 +8,7 @@ export class TierService {
 
   constructor(private configService: ConfigService) {}
 
-  async checkAddOrg(userId: number, userDetails: { email: string; name: string; login: string }, proAccount = false) {
+  async checkAddOrg(userId: number, userDetails: { email: string; name: string; login: string }) {
     try {
       // check to see if the user is registered
       await lookupOrg(`org:${userId}`);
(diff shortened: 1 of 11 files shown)


src/tier/tier.service.ts
```diff
@@ -0,0 +1,41 @@
...
+            description: `OpenSauced Free Account`,
+            phone: "",
+            metadata: {
+              login,
+            },
+          },
+        });
+      } catch (e) {
+        this.logger.error(`Unable to register account for ${login}`);
+      }
```
jpmcb, line 38: Do we want to surface the error here? How we've done error handling in the past has confused me abit due to typescripts gnarly handling of errors but elsewhere I've seen us do:

https://github.com/open-sauced/api/blob/692f953192dff8449f3716a0a28c88f20f1f741c/src/blogs/blog-summary.service.ts#L31-L34

src/user/services/user.service.ts
```diff
@@ -261,6 +263,12 @@ export class UserService {
         });
       }
 
+      await this.tierService.checkAddOrg(
+        id,
+        { email: email as string, name: name as string, login: user_name as string },
+        user.role >= 50
```
jpmcb, line 269: So "user.role greater than or equal to 50" is how we grandfather legacy privileged accounts into the tier.run system?

jpmcb (approved):

src/user/services/user.service.ts
brandonroberts: Yes, but on second thought we can do that outside of the API and leave this to the free tier.

src/tier/tier.service.ts
brandonroberts: Yep, good catch

PR #444, opened 2023-12-11
feat: V2 pr search using timescale data

jpmcb:
## Description

This is a deep refactor that adds a v2 endpoint for `prs/search`.

This should capture all the functionality of the existing `prs/search` except for searching for spam PRs. There's some missing data in the timescale ingestion that we could look at building into the dataset

- [x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI
- [ ] 📦 Chore (Release)
- [ ] ⏩ Revert

Related to:
- https://github.com/open-sauced/api/pull/442
- https://github.com/open-sauced/app/pull/2293

## Steps to QA

Run locally, connect to a timescale instance, make some queries via the swagger:

Screenshot 2023-12-11 at 4 48 20 PM

## Added to documentation?

- [ ] 📜 README.md
- [ ] 📓 docs.opensauced.pizza
- [ ] 🍕 dev.to/opensauced
- [ ] 📕 storybook

### DIFF
--- a/src/timescale/pull_request_github_events.service.ts
+++ b/src/timescale/pull_request_github_events.service.ts
@@ -1,17 +1,26 @@
 import { Injectable } from "@nestjs/common";
 import { InjectRepository } from "@nestjs/typeorm";
-import { Repository } from "typeorm";
+import { Repository, SelectQueryBuilder } from "typeorm";
+import { ConfigService } from "@nestjs/config";
+import { FilterListContributorsDto } from "../user-lists/dtos/filter-contributors.dto";
+import { RepoService } from "../repo/repo.service";
+import { PullRequestPageOptionsDto } from "../pull-requests/dtos/pull-request-page-options.dto";
+import { RepoSearchOptionsDto } from "../repo/dtos/repo-search-options.dto";
 import { PageOptionsDto } from "../common/dtos/page-options.dto";
 import { PageMetaDto } from "../common/dtos/page-meta.dto";
 import { PageDto } from "../common/dtos/page.dto";
 import { GetPrevDateISOString } from "../common/util/datetimes";
+import { UserListService } from "../user-lists/user-list.service";
 import { DbPullRequestGitHubEvents } from "./entities/pull_request_github_event";
 
 @Injectable()
 export class PullRequestGithubEventsService {
   constructor(
     @InjectRepository(DbPullRequestGitHubEvents, "TimescaleConnection")
-    private pullRequestGithubEventsRepository: Repository<DbPullRequestGitHubEvents>
+    private pullRequestGithubEventsRepository: Repository<DbPullRequestGitHubEvents>,
+    private readonly configService: ConfigService,
+    private readonly repoService: RepoService,
+    private readonly userListService: UserListService
   ) {}
 
   baseQueryBuilder() {
@@ -20,26 +29,15 @@ export class PullRequestGithubEventsService {
     return builder;
   }
 
-  async findAllByPrAuthor(author: string, pageOptionsDto: PageOptionsDto): Promise<PageDto<DbPullRequestGitHubEvents>> {
-    const startDate = GetPrevDateISOString(pageOptionsDto.prev_days_start_date);
-    const range = pageOptionsDto.range!;
-    const order = pageOptionsDto.orderDirection!;
-
-    /*
-     * because PR events may be "opened" or "closed" many times, this inner CTE query gets similar PRs rows
-     *  based on pr_number and repo_name. This essentially gives a full picture of opened/closed PRs
-     *  and their current state
-     */
-
-    const cteBuilder = this.pullRequestGithubEventsRepository
-      .createQueryBuilder("pull_request_github_events")
-      .select("*")
-      .addSelect(`ROW_NUMBER() OVER (PARTITION BY pr_number, repo_name ORDER BY event_time ${order}) AS row_num`)
-      .where(`"pull_request_github_events"."pr_author_login"=:author`, { author: author.toLowerCase() })
-      .andWhere(`'${startDate}'::TIMESTAMP >= "pull_request_github_events"."event_time"`)
-      .andWhere(`'${startDate}'::TIMESTAMP - INTERVAL '${range} days' <= "pull_request_github_events"."event_time"`)
-      .orderBy("event_time", order);
-
+  /*
+   * this function takes a cte builder and gets the common rows for pull_request_github_events
+   * off of it. It also builds a cte counter to ensure metadata is built correctly
+   * for the timescale query.
+   */
+  async execCommonTableExpression(
+    pageOptionsDto: PageOptionsDto,
+    cteBuilder: SelectQueryBuilder<DbPullRequestGitHubEvents>
+  ) {
     const queryBuilder = this.pullRequestGithubEventsRepository.manager
       .createQueryBuilder()
       .addCommonTableExpression(cteBuilder, "CTE")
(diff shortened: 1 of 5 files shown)


jpmcb:
We'll want to figure out how to handle spam PRs right now: where is that getting surface in the app right now? We could always build out some custom frontend stuff for that. I'm not sure if the existing spam queries are going through the `v1/prs/search` endpoint

src/timescale/pull_request_github_events.service.ts
```diff
@@ -91,4 +89,156 @@ export class PullRequestGithubEventsService {
...
+
+    /*
+     * because PR events may be "opened" or "closed" many times, this inner CTE query gets similar PRs rows
+     * based on pr_number and repo_name. This essentially gives a full picture of opened/closed PRs
+     * and their current state
+     */
+
+    const cteBuilder = this.pullRequestGithubEventsRepository
+      .createQueryBuilder("pull_request_github_events")
+      .select("*")
```
nickytonline, line 131: Is `*` what we want or do we plan to be more specific here?

src/timescale/pull_request_github_events.service.ts
```diff
@@ -91,4 +89,156 @@ export class PullRequestGithubEventsService {
...
+          .where(`'${startDate}'::TIMESTAMP >= "pull_request_github_events"."event_time"`)
+          .andWhere(
+            `'${startDate}'::TIMESTAMP - INTERVAL '${range} days' <= "pull_request_github_events"."event_time"`
+          );
+        break;
+    }
+
+    /* filter on PR author / contributor */
+    if (pageOptionsDto.contributor) {
+      cteBuilder.andWhere(`"pull_request_github_events"."pr_author_login"=:author`, {
```
brandonroberts, line 177: If this is case-senstive, we should lowercase both values for the `pr_author_login=:author`

src/timescale/pull_request_github_events.service.ts
```diff
@@ -91,4 +89,156 @@ export class PullRequestGithubEventsService {
...
+      cteBuilder.andWhere(`"pull_request_github_events"."repo_name" IN (:...repoNames)`, {
+        repoNames,
+      });
+    }
+
+    /* apply user provided repo name filters */
+    if (pageOptionsDto.repo) {
+      cteBuilder.andWhere(`"pull_request_github_events"."repo_name" IN (:...repoNames)`, {
+        repoNames: pageOptionsDto.repo.split(","),
+      });
```
brandonroberts, line 208: This should be just one unless we are building in support to expand it into multiples for later.

```suggestion
      cteBuilder.andWhere(`"pull_request_github_events"."repo_name" =:repo`, {
        repoNames: pageOptionsDto.repo,
      });
```

src/timescale/pull_request_github_events.service.ts
jpmcb: Great question: this is _abit_ nuanced since we're using conman table expressions without typeorm handling the actual table relations for us.

This essentially gets a CTE of all the values in the `pull_request_github_events` given the `where` constraints below and partitions each in the common table based on their pr_number and repo_name. We need the CTE to give us this partitioning of prs since there may be many different events for a single PR (opening, closing, etc.).

This then is further filtered out by the actual `WITH` which uses the CTE and that gets the specific rows we care about (which is all abstracted by `.addCommonTableExpression`). These results are then cast to the right type in the new `execCommonTableExpression` function. Since both `findAllByPrAuthor` and the new `findAllWithFilters` used this same CTE builder method, it made sense to abstract that function out.

The end resulting compiled SQL looks like this:
```sql
WITH "CTE" AS (
  -- get common, partitioned rows of PR events partitioned by pr / repo name
	SELECT
		*,
		ROW_NUMBER() OVER (PARTITION BY pr_number,
			repo_name ORDER BY event_time DESC) AS row_num
	FROM
		"pull_request_github_events" "pull_request_github_events"
	WHERE
		'2023-12-11T23:22:51.163Z'::TIMESTAMP >= "pull_request_github_events"."event_time"
		AND '2023-12-11T23:22:51.163Z'::TIMESTAMP - INTERVAL '30 days' <= "pull_request_github_events"."event_time"
   -- additional WHERE constraints
		ORDER BY
			event_time DESC
)
SELECT
	event_id,
  pr_number,
  -- ... select for additional rows of data we care about 
FROM
	"CTE" "CTE"
WHERE
	row_num = 1
LIMIT 10;
```

To play devils advocate: we could flip this around and select for rows we care about in the CTE and select all those rows with `*` in the actual `WITH / SELECT / FROM` which should reduce the amount of cruft flowing through.

src/timescale/pull_request_github_events.service.ts
```diff
@@ -91,4 +89,156 @@ export class PullRequestGithubEventsService {
...
+      .select("*")
+      .addSelect(`ROW_NUMBER() OVER (PARTITION BY pr_number, repo_name ORDER BY event_time ${order}) AS row_num`)
+      .orderBy("event_time", order);
+
+    /*
+     * todo: filter for spam PRs.
+     * The way this was done in v1 was to check for any and all PRs in repos
+     * that had a "spam" like label on them. As of this writing, the events feed
+     * nozzle does not capture labels for PRs / issues.
+     */
```
jpmcb, line 157: Just want to call this out: I'm not seeing a clear path with the data right now to filter for spam PRs. It seems the way this was done in v1 was basically to select for PRs that have a `spam` label. The timescale data, at least right now, does not have label data. We'd need to start consuming that from the nozzle to get this data to flow in.

src/timescale/pull_request_github_events.service.ts
jpmcb: I'm hesitant to apply `LOWER` here since that will negate indexes we have on `pr_author_login` which were created without `LOWER` on usernames. This will be a big hit to performance.

To my understanding, GitHub logins are _not_ case-sensitive, at least from GitHub's side. For example, https://github.com/BrandonRoberts is no different from https://github.com/brandonroberts

jpmcb: But I suppose that's a pretty bad user experience if someone is entering `BrandonRoberts` since the index will be on `brandonroberts` ... I suppose we should probably go back and rebuild indexes based on the `LOWER` of all usernames.

src/timescale/pull_request_github_events.service.ts
jpmcb: This definitely works for multiple repo names. For example:

```
curl -X 'GET' \
  'http://localhost:3003/v2/prs/search?page=1&limit=10&range=30&prev_days_start_date=0&repo=open-sauced%2Fapp%2Copen-sauced%2Fapi' \
  -H 'accept: application/json'
```

which sets repo is set as:
```
open-sauced/app,open-sauced/api
```

returns results for both repos.

src/timescale/pull_request_github_events.service.ts
bdougie: We can remove this. Almost nobody uses that label outside of hacktoberfest. We should remove the label filtering for now and think of a better solution in the summer. 

My recommendation for addressing the Spam is to use ML model in the future. Identify what spam typically looks like and have the user mark that in OpenSauced, not GitHub. But let's save this conversation for a later date.

jpmcb:
The `LOWER` indexes have landed on the database and this PR has been updated with relevant usage of those indexes!

This can go in anytime as I shift to wrapping up the integration in: https://github.com/open-sauced/app/pull/2293

nickytonline (approved):
:shipit:

brandonroberts (approved):
🍕

Issue #478, opened 2024-01-10
Annoucement 🚨 `v2` breaking changes 


jpmcb:
In an effort to migrate away from data scrapped off of GitHub’s API and move to our new events based data stored in Timescale, OpenSauced engineering will be cutting a new, breaking `v2` of the OpenSauced API.

Given there are few direct consumers of the API outside of the `open-sauced` org, there will be _no transition period between the `v1` and `v2` API_. Client consumers (like `open-sauced/app`, `open-sauced/hot`, `open-sauced/pizza-cli`, etc.) should anticipate these breaking changes and release concurrently with beta and prod releases of the API.

So, for example:

```
GET v1/auth/session
```

will become

```
GET v2/auth/session
```

regardless of functionality change.

---

### But wait! I consume the OpenSauced API!

Fabulous! We'd love to learn more about your use case, how you consume the OpenSauced API, and how we can work with you to make this a smooth transition for you! In general, most routes will not change and you should be ok with removing/replacing `v1` with `v2`.

### When will this happen?

We will keep this issue up to date on when this will happen but we will eventually cut these changes into `beta` from `alpha` and then into `main` from `beta`. Keep an eye on those branches to stay in the loop.

### What's changing in the API?

All routes will now start with `v2`, there will be a few new routes, some data will be mutated in a few routes, and several routes are being removed.

This issue will be kept up to date with what's happening as we complete this work.

bdougie:
Thanks for this @jpmcb. Commenting to subscribe and ensure I create a transition plan for the following:

- [ ] open-sauced/hot
- [ ] open-sauced/ai
- [ ] open-sauced/contributor.info

jpmcb:
V2 has been released 🐙 https://github.com/open-sauced/api/pull/524

PR #485, opened 2024-01-13
feat: v2 API 👨🏻‍🍳 - V2 contrib evolution

jpmcb:
## Description

This implements a much faster v2 contributor evolution histogram:

---

- Updates the missing v2 for swagger generation
- Updates the `DbContributionStatTimeframe` type to have all the new data points
- Implements a `findContributorsByType` helper function which returns usernames for the given "type" (churn, new, alum, etc.) This was _much_ easier than trying to implement this across a `time_bucket` which I'm not sure is possible. Or at least, it'd be very gross to try to do without a helper query/function
- Implements `findContributionsInTimeFrame` which uses `UNION ALL` across many of the events tables given the usernames

Example `v2/lists/{id}/stats/contributions-evolution-by-type`:

```json5
[
  {
    "bucket": "2024-01-08T00:00:00.000Z",
    "commits": 19,
    "prs_created": 1,
    "prs_reviewed": 17,
    "issues_created": 6,
    "commit_comments": 0,
    "issue_comments": 17,
    "pr_review_comments": 5,
    "comments": 22,
    "total_contributions": 65
  },

  // ... redacted

  {
    "bucket": "2023-12-14T00:00:00.000Z",
    "commits": 3,
    "prs_created": 0,
    "prs_reviewed": 9,
    "issues_created": 1,
    "commit_comments": 0,
    "issue_comments": 4,
    "pr_review_comments": 4,
    "comments": 8,
    "total_contributions": 21
  }
]
```

- [x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI
- [ ] 📦 Chore (Release)
- [ ] ⏩ Revert

Related to v2 changes:
- https://github.com/open-sauced/engineering/issues/217
- https://github.com/open-sauced/api/issues/478

## Added to documentation?

- [ ] 📜 README.md
- [ ] 📓 docs.opensauced.pizza
- [ ] 🍕 dev.to/opensauced
- [ ] 📕 storybook

### DIFF
--- a/src/user-lists/entities/contributions-timeframe.entity.ts
+++ b/src/user-lists/entities/contributions-timeframe.entity.ts
@@ -4,8 +4,8 @@ import { Column, Entity } from "typeorm";
 @Entity({ name: "user_list_contributors" })
 export class DbContributionStatTimeframe {
   @ApiModelProperty({
-    description: "The ISO timestamp for the start of the time frame",
-    example: "2023-08-26T23:55:49.204Z",
+    description: "The ISO timestamp for the given time bucket",
+    example: "2024-01-08T00:00:00.000Z",
     type: "string",
   })
   @Column({
@@ -13,22 +13,10 @@ export class DbContributionStatTimeframe {
     select: false,
     insert: false,
   })
-  time_start: string;
+  bucket: string;
 
   @ApiModelProperty({
-    description: "The ISO timestamp for the end of the time frame",
-    example: "2023-08-26T23:55:49.204Z",
-    type: "string",
-  })
-  @Column({
-    type: "string",
-    select: false,
-    insert: false,
-  })
-  time_end: string;
-
-  @ApiModelProperty({
-    description: "Number of commits associated with a user login",
+    description: "Number of commits within the time range",
     example: 0,
     type: "integer",
   })
@@ -40,7 +28,7 @@ export class DbContributionStatTimeframe {
   commits: number;
 
   @ApiModelProperty({
-    description: "Number of PRs created associated with a user login",
+    description: "Number of PRs created within the time range",
     example: 0,
     type: "integer",
   })
@@ -52,7 +40,7 @@ export class DbContributionStatTimeframe {
   prs_created: number;
 
   @ApiModelProperty({
-    description: "Number of PRs reviewed by a user login",
+    description: "Number of PRs reviewed within the time range",
     example: 0,
     type: "integer",
   })
@@ -64,7 +52,7 @@ export class DbContributionStatTimeframe {
   prs_reviewed: number;
 
   @ApiModelProperty({
-    description: "Number of issues created by a user login",
+    description: "Number of issues within the time range",
     example: 0,
     type: "integer",
   })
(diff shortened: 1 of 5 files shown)


src/user-lists/entities/contributions-timeframe.entity.ts
```diff
@@ -75,6 +63,42 @@ export class DbContributionStatTimeframe {
   })
   issues_created: number;
 
+  @ApiModelProperty({
+    description: "",
```
jpmcb, line 67: TODO: These need descriptions

src/user-lists/user-list-events-stats.service.ts
```diff
@@ -222,6 +295,103 @@ export class UserListEventsStatsService {
...
+    if (allUsers.length === 0) {
+      return new Array<DbContributionStatTimeframe>();
+    }
+
+    const usersWhereConditions = allUsers.map((user) => `LOWER(actor_login) = LOWER('${user}')`).join(" OR ");
+
+    const cteQuery = `
+      SELECT event_type, event_time
+      FROM push_github_events
+      WHERE (${usersWhereConditions})
```
jpmcb, line 316: There's probably a better way to do this with the CTE and using the `setParameters` function

jpmcb (commented):
Marking this as draft for now and coming back to it after the weekend

jpmcb:
This is ready to rock and roll into `alpha`: force pushed to add descriptions to the virtual entity and addressed using parameters instead of raw `user` strings stuck together in the `WHERE` for the union tables.

src/user-lists/entities/contributions-timeframe.entity.ts
```diff
@@ -4,31 +4,31 @@ import { Column, Entity } from "typeorm";
 @Entity({ name: "user_list_contributors" })
 export class DbContributionStatTimeframe {
   @ApiModelProperty({
-    description: "The ISO timestamp for the start of the time frame",
-    example: "2023-08-26T23:55:49.204Z",
+    description: "The ISO timestamp for the given time bucket",
```
nickytonline, line 7: Curious about the wording change. Doesn't it mean the same thing and a time bucket is the implementation of a time frame?

src/user-lists/user-list-events-stats.service.ts
```diff
@@ -39,6 +41,77 @@ export class UserListEventsStatsService {
...
+    const entityQb = this.pullRequestGithubEventsRepository.manager
+      .createQueryBuilder()
+      .addCommonTableExpression(userListQueryBuilder, "CTE")
+      .setParameters(userListQueryBuilder.getParameters())
+      .select("login")
+      .from("CTE", "CTE");
+
+    const entities = await entityQb.getRawMany<{ login: string }>();
+
+    return entities.map((result) => result.login);
```
nickytonline, line 112: IS the map necessary if we're getting back an array of `{ login: string }`? or is `getRawMany` returning more than logins?

src/user-lists/entities/contributions-timeframe.entity.ts
jpmcb: It is different: the `time_bucket` function in timescale doesn't return the iso timestamp for the _start of the timebucket_, it uses a more complex "origin" system that determines the start of the bucket on sliding windows from the first possible timestamped data.

> By default, a time bucket doesn't start at the earliest timestamp in your data. There is often a more logical time ... time is divided into buckets based on intervals from the origin.

https://docs.timescale.com/use-timescale/latest/time-buckets/about-time-buckets/ 

TLDR: Timescale does intelligent bucketing based on what data it is presented with. These are not necessarily the beginnings of the "data" per-say, which is why you might see drifts from what range you originally provided, but rather, the start of the bucket for what logical data fits into those buckets.

I'm honestly not sure I fully understand it yet, but this is a good thing for us not having to create complex software (like we had before) that looks at data and tries to create buckets based on it.

---

I can go back on this documentation but it made sense to remove the notion of "beginning" since that seemed more complex than just calling it the timestamp of the bucket.

src/user-lists/user-list-events-stats.service.ts
jpmcb: Yeah, I originally tried that, but `getRawMany` with the type cast creates an array of objects whose keys are `login` - this is needed since we are selecting for the `login` column in the query.

I didn't want to have to export that `{ login: string }` type since it's just a one off to get an array of users. So it seemed simpler to just map it out to an array.

nickytonline (approved):
:shipit:

brandonroberts (approved):

PR #533, opened 2024-02-01
fix: user-list stats queries ignore empty usernames

jpmcb:
## Description

- This fixes an edge case where a user list may have empty usernames. If present, those empty usernames are filtered out.
- Removes unused user-list-stats service (this was replaced by the user list events service)
- Removes unused "login" field in entity (the `user_list_contributors` table only has "username")

## What type of PR is this? (check all applicable)

- [ ] 🍕 Feature
- [x] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] 🔥 Performance Improvements
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI
- [ ] 📦 Chore (Release)
- [ ] ⏩ Revert

## Related Tickets & Documents

N/a - followup to #529 

## Steps to QA

Call `most-active-contributors` with a very big list of users:

Screenshot 2024-02-01 at 10 22 03 AM

Returns very quickly and does not brick the database

## Added to documentation?

- [ ] 📜 README.md
- [ ] 📓 docs.opensauced.pizza
- [ ] 🍕 dev.to/opensauced
- [ ] 📕 storybook
- [ ] 🙅 no documentation needed

PR #538, opened 2024-02-02
fix: add login to user list contributor data

brandonroberts:
## Description

- Adds the `login` field back to the user list contributor entity as needed for the UI. We left join on the users table, and the login is empty if the user_id is not in our data, which would fallback to the `username`.

- [ ] 🍕 Feature
- [x] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI
- [ ] 📦 Chore (Release)
- [ ] ⏩ Revert

BEFORE:

image

AFTER:

image

## Added to documentation?

- [ ] 📜 README.md
- [ ] 📓 docs.opensauced.pizza
- [ ] 🍕 dev.to/opensauced
- [ ] 📕 storybook

### DIFF
--- a/src/user-lists/entities/user-list-contributor.entity.ts
+++ b/src/user-lists/entities/user-list-contributor.entity.ts
@@ -72,4 +72,15 @@ export class DbUserListContributor {
     referencedColumnName: "id",
   })
   public user_list_contributor!: DbUser;
+
+  @ApiModelProperty({
+    description: "User list collaborator's login",
+    example: "bdougie",
+  })
+  @Column({
+    type: "text",
+    select: false,
+    insert: false,
+  })
+  public login?: string;
 }


nickytonline (approved):
🚢

jpmcb (approved):
This LGTM but I'm abit confused why this is needed - I see we are left joining on the `users` table here:

https://github.com/open-sauced/api/blob/123b94b3b7ec2a87147b4bf370d4b5f37c0b4ff3/src/user-lists/user-list.service.ts#L219-L240

which gets used in the user list `:id/contributors` endpoint and would populate a `login` from the `users` table. Can we not reliably use `username` in the `user_list_contributors` table? Or are we not always assured `username` will be filled? This is probably also related to https://github.com/open-sauced/etl/issues/99 and I'm probably getting confused on the "contract" around `username` in this table.

I had originally removed this in https://github.com/open-sauced/api/pull/533 seeing that `login` was not referenced anywhere via the typescript LSP (but alas, i should have known better and should have looked if we were populating that entity's field anywhere in a query) and assumed that `username` would always be there.

Do we need to go back in that query for the most-active-contributor and attempt to fallback on `login` as well?

Issue #561, opened 2024-02-13
feat: Allow for contributors to be added during workspace creation


jpmcb:
Currently, we only allow for `repos` during workspace creation:

https://github.com/open-sauced/api/blob/7c4d7f8d0aacc5fabf8fb7f1e0326632bfb3e261/src/workspace/dtos/create-workspace.dto.ts#L39-L46

We should also add a `contributors` key that is an object of ids/logins

---

cc @zeucapua

PR #568, opened 2024-02-14
feat: Workspace creation supports contributors

jpmcb:
## Description

This patch supports a list of contributors for workspace creation.

- [x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI
- [ ] 📦 Chore (Release)
- [ ] ⏩ Revert

Closes: https://github.com/open-sauced/api/issues/561

Related to: https://github.com/open-sauced/app/pull/2658

cc @zeucapua 

## Steps to QA

1. Run API locally attached to database
2. Create a new workspace with a body like:

```sh
curl -X 'POST' \
  'http://localhost:3003/v2/workspaces' \
  -H 'Authorization: Bearer <redacted>' \
  -H 'Content-Type: application/json' \
  -d '{
  "name": "My Workspace",
  "description": "A workspace for my OSS collaborators",
  "members": [],
  "repos": [
    {
      "full_name": "open-sauced/api"
    }
  ],
  "contributors": [
    {
      "login": "bdougie"
    }
  ]
}'
```

and the workspace is successfully created with associated repo/contributor from list.

## Added to documentation?

- [ ] 📜 README.md
- [ ] 📓 docs.opensauced.pizza
- [ ] 🍕 dev.to/opensauced
- [ ] 📕 storybook

### DIFF
--- a/src/user/services/user.service.ts
+++ b/src/user/services/user.service.ts
@@ -1,4 +1,4 @@
-import { BadRequestException, Injectable, NotFoundException } from "@nestjs/common";
+import { BadRequestException, Inject, Injectable, NotFoundException, forwardRef } from "@nestjs/common";
 import { Repository, SelectQueryBuilder } from "typeorm";
 import { InjectRepository } from "@nestjs/typeorm";
 import { User } from "@supabase/supabase-js";
@@ -43,6 +43,7 @@ export class UserService {
     private workspaceRepository: Repository<DbWorkspace>,
     @InjectRepository(DbWorkspaceMember, "ApiConnection")
     private workspaceMemberRepository: Repository<DbWorkspaceMember>,
+    @Inject(forwardRef(() => PullRequestGithubEventsService))
     private pullRequestGithubEventsService: PullRequestGithubEventsService
   ) {}
 
--- a/src/workspace/workspace.service.ts
+++ b/src/workspace/workspace.service.ts
@@ -1,4 +1,4 @@
-import { Injectable, NotFoundException, UnauthorizedException } from "@nestjs/common";
+import { BadRequestException, Injectable, NotFoundException, UnauthorizedException } from "@nestjs/common";
 import { Repository, SelectQueryBuilder } from "typeorm";
 import { InjectRepository } from "@nestjs/typeorm";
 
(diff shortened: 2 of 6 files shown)


src/user/services/user.service.ts
```diff
@@ -43,6 +43,7 @@ export class UserService {
     private workspaceRepository: Repository<DbWorkspace>,
     @InjectRepository(DbWorkspaceMember, "ApiConnection")
     private workspaceMemberRepository: Repository<DbWorkspaceMember>,
+    @Inject(forwardRef(() => PullRequestGithubEventsService))
```
jpmcb, line 46: The `UserModule` keeps complaining about this circular dependency. Which is strange since this patch doesn't add any additional dependencies to any modules. Maybe this is what it needed in the end, but either way, I messed around with some of the deps/forward refs and it's happy now.

brandonroberts (approved):

src/workspace/workspace.service.ts
```diff
@@ -200,6 +205,43 @@ export class WorkspaceService {
...
+          const newContributor = new DbWorkspaceContributor();
+
+          newContributor.workspace = savedWorkspace;
+          newContributor.contributor = user;
+
+          await entityManager.save(DbWorkspaceContributor, newContributor);
+        }
+      });
+
+      await Promise.all(contribPromises);
```
brandonroberts, line 243: Do we want to use `Promise.allSettled` here instead so one failure doesn't cause a completely failed request? I could see us running into failures of not having users in our table.

jpmcb: Hmmm good question: this entire block is wrapped in a `workspaceRepository.manager.transaction` which will intentially rollback the entire workspace creation transaction if there are any failures (so same scenario for repos we're missing).

That's what I plan to tackle next: make users / repos more agnostic to be backfilled by the ETL latter.

But for now, I think this works and we go back and refactor that in a future fix as the ETL will be fixed to work with it all

brandonroberts: Ahh ok, I missed the transaction block 👍

nickytonline (approved):
🚢

Issue #511, opened 2024-01-24
Bug: adding workspace repos fails if we are missing repo data


jpmcb:
### Describe the bug

Given a valid repo (although abit bespoke and untouched) `jpmcb/go-interview-manual`, creating / adding repos to a workspace fails as if the repo did not exist:

```
{
  "statusCode": 404,
  "message": "Repository not found: jpmcb/go-interview-manual",
  "error": "Not Found"
}
```

This is because our data does not have that repo:

```sql
select * from repos where full_name = 'jpmcb/go-interview-manual';

-- yields nothing
```

### Steps to reproduce

1. Find a repo that is not in our `repos` table
2. Attempt to add it as a repo to a workspace
3. Notice the failure

---

There are a few things going on here: our ETL does not scrape _all_ repos known to GitHub (since many are defunct, too small to care about, or 

But with workspaces, a user can search through GitHub for repos to add to a workspace (see https://github.com/open-sauced/app/pull/2502) which may lead to situations where they want to add a repo to a workspace, we don't have it in our data, and it fails.

### Possible solutions

We could add a mechanism that adds an "empty" repo row in the `repos` table if we don't have it when a user is attempting to add repos to workspaces. This way, the ETL will pick it up pretty quickly (as long as we mark the `updated_at` field as `1970...`)

But this _may_ come with a few problems we want to consider:
- The API currently has no GitHub validation: i.e., a client could attempt to add a repo that does not exist, it gets added as an empty row and never gets updated. We'd end up with abunch of junk rows and in a worse case scenerio, a malicious actor blasts us with junk data that fills up and bricks our database. Not ideal! We'd likely want to either have API validation that the repo is really a real GitHub repo if it can't find it in the `repo` table and if the ETL can't find it, the ETL should go ahead and remove it.
  - This "cleanup" in itself presents some additional unique problems: the ETL would probably need to also look for "orphaned" workspace_repo join rows that don't get automatically deleted when a `repos` row is cleaned up.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

### Contributing Docs

- [X] I agree to follow this project's Contribution Docs

PR #573, opened 2024-02-15
feat: Add octokit and fetching users/repos for workspaces flows

jpmcb:
## Description

This adds octokit (alongside a personal web token auth method) to fetch github users / repos when workspaces flows add unknown users/repos.

The biggest thing is that we need the user/repo ID: because we don't generate IDs as the primary key for either the user table nor the repo table, we get those directly from GitHub. So attempting to come up with a "temporary" or partial entry that can latter get picked up by the ETL is abit of a mess.

Instead, this uses the octokit directly and gets those entities and fills in the necessary partial entries to be filled in at a latter time by the ETL.

---

This also refactors the workspaces repos and contributors service functions to use a common "execute..." function that performs that functionality.

- [x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI
- [ ] 📦 Chore (Release)
- [ ] ⏩ Revert

Closes: https://github.com/open-sauced/api/issues/511

## Steps to QA

TODO - still doing some deeper validation with the ETL too.

## Added to documentation?

- [ ] 📜 README.md
- [ ] 📓 docs.opensauced.pizza
- [ ] 🍕 dev.to/opensauced
- [ ] 📕 storybook

### DIFF
--- a/src/insight/dtos/create-insight.dto.ts
+++ b/src/insight/dtos/create-insight.dto.ts
@@ -25,7 +25,7 @@ export class CreateInsightDto {
     description: "An array of repository information objects",
     isArray: true,
     type: RepoInfo,
-    example: [{ id: 797, full_name: "open-sauced/insights" }],
+    example: [{ id: 797, fullName: "open-sauced/insights" }],
   })
   @Type(() => RepoInfo)
   @IsArray()
--- a/migrations/20240215112913_user_fetch_error_column.sql
+++ b/migrations/20240215112913_user_fetch_error_column.sql
@@ -1,0 +1,16 @@
+-------------------------------------------------------------------------------
+-- This adds a "fetch_error" column to the users and repos tables: this is in service
+-- of surfacing issues with the ETL not being able to fetch data from GitHub on entities.
+-------------------------------------------------------------------------------
+
+-- Users fetch_error column
+alter table users
+  add column fetch_error boolean default false;
+
+create index idx_users_fetch_error on users (fetch_error);
+
+-- Repos fetch_error column
+alter table repos
+  add column fetch_error boolean default false;
+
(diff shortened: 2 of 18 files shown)


jpmcb:
cc @open-sauced/engineering  - opening this now to get some validation this is the right approach: we'll need another github auth token or spend the time getting this integrated with a github org app.

jpmcb:
Force pushed to add some additional checks on the repo stub creation: now, users (which get referenced in the foreign key `user_id`) get created correctly as well.

src/insight/dtos/create-insight.dto.ts
```diff
@@ -25,7 +25,7 @@ export class CreateInsightDto {
     description: "An array of repository information objects",
     isArray: true,
     type: RepoInfo,
-    example: [{ id: 797, full_name: "open-sauced/insights" }],
+    example: [{ id: 797, fullName: "open-sauced/insights" }],
```
nickytonline, line 28: is this change because we're bringnig in Octokit, so JS camelCase?

jpmcb: This was a typo in the swagger stuff: the field for that DTO isn't `full_name`, it's `fullName` which was erroring out when trying to use the swagger UI. So this is just a one off typo fix 😅

nickytonline: 😅

nickytonline (approved):
:shipit:

brandonroberts (approved):

PR #576, opened 2024-02-16
feat: Supabase passthrough auth

jpmcb:

- [ ] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI
- [ ] 📦 Chore (Release)
- [ ] ⏩ Revert

## Added to documentation?

- [ ] 📜 README.md
- [ ] 📓 docs.opensauced.pizza
- [ ] 🍕 dev.to/opensauced
- [ ] 📕 storybook


jpmcb:
> the incoming @UserId() potentially be undefined? , correct?

That's correct: the userId in those cases would potentially be undefined. Maybe I add another decorator, like `@OptionalUserId()` or something, in order to reduce the blast radius of this.

brandonroberts:
> > the incoming @userid() potentially be undefined? , correct?
> 
> That's correct: the userId in those cases would potentially be undefined. Maybe I add another decorator, like `@OptionalUserId()` or something, in order to reduce the blast radius of this.

👍

src/stripe/stripe.service.ts
```diff
@@ -43,4 +43,23 @@ export class StripeService {
...
+          quantity: 1,
+        },
+      ],
+      mode: "subscription",
+      success_url: `${this.configService.get<string>("stripe.subscriptionSessionCheckoutSuccessURL")!}`,
+      cancel_url: `${this.configService.get<string>("stripe.subscriptionSessionCancelURL")!}`,
+    });
+
+    return { sessionId: session.id };
+  }
```
jpmcb, line 64: cc @brandonroberts - I'm unsure if this is the right approach or if you can point me to something better. But this all essentially assumes we'll create a separate session with a separate configured price/etc.

jpmcb:
Force pushed to add some pricing stubs: this still needs work and I'm blocked on testing this. But in short this adds:

- A `payee_customer_stripe_id` column to workspaces that references a stripe customer id in the `customers` table. This will be used to determine if the workspace can be made private.
- A `/checkout/workspaces/:id/session` route that can be used to create a new stripe session for a certain workspace
  - A few stubs for creating the session, etc. It's still known how this will all work without testing it / bootstrapping it in the stripe dashboard.

brandonroberts:
@jpmcb this looks good to me. We can test the workflow on beta after we get something on the UI side

migrations/20240216150233_public_workspaces.sql
```diff
@@ -0,0 +1,22 @@
...
+--
+-- The workspaces.payee_customer_stripe_id column is a foreign key reference to a paying
+-- user which denotes the subscription for the paid workspace.
+-------------------------------------------------------------------------------
+
+alter table workspaces
+  add column is_public boolean default true;
+
+alter table workspaces
+  add column payee_customer_stripe_id text default null;
```
jpmcb, line 17: Whoops, TODO - this is old, delete this column

nickytonline (approved):

Issue #592, opened 2024-02-22
feat: Revive hot.opensauced.pizza with V2 data


bdougie:
Got some feedback from an open source maintainer asking if we share a list of most starred or commented projects per day. My response was we did, and it was hot.opensauced.pizza. 

## Proposal
I'd like to propose adding some data flexibility to the histogram so we can do the following on hot.

- List most starred projects per day (Popular)
- List most forked projects per day (Forked)
- List most commented projects per day (Discussed)
- List most starred repos created in the last 30 days (New).
- I am not sure voting is needed, but we could keep it since it already exists.

This would align with the existing filters, but provide better data.

![Screen Shot 2024-02-22 at 1 58 42 PM](https://github.com/open-sauced/api/assets/5713670/2e0d2f01-b5d0-4d2c-ab6d-9c45dbe8cc50)

## SQL query

I played with the SQL query, but it still needs work. Sharing now so I can move on to something more pressing

```sql
SELECT
  repo_name, 
  time_bucket('1 day', event_time) AS bucket,
  count(*) as star_count
FROM watch_github_events
WHERE event_time > now() - INTERVAL '1 days'
GROUP BY repo_name, event_time
ORDER BY star_count desc;
```
I found that the query above was providing duplicates

![Screen Shot 2024-02-22 at 2 01 44 PM](https://github.com/open-sauced/api/assets/5713670/8bdcf2ea-1df6-4925-a127-fa8f322bd2f1)

PR #597, opened 2024-02-23
feat: Implements top new starred repos, forks, and comments

jpmcb:
## Description

This implements:
- Top forked repos

Screenshot 2024-02-23 at 12 20 44 PM

- Top commented repos: this incorporates both issue comments and pr review comments

Screenshot 2024-02-23 at 12 21 18 PM

- Top _new_ starred repos: this filters down the top repo results down to only those that are newly created/newly made public

Screenshot 2024-02-23 at 12 22 09 PM

- [x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI
- [ ] 📦 Chore (Release)
- [ ] ⏩ Revert

Closes: https://github.com/open-sauced/api/issues/592

## Added to documentation?

- [ ] 📜 README.md
- [ ] 📓 docs.opensauced.pizza
- [ ] 🍕 dev.to/opensauced
- [ ] 📕 storybook

### DIFF
--- a/src/timescale/fork_github_events.service.ts
+++ b/src/timescale/fork_github_events.service.ts
@@ -2,7 +2,7 @@ import { BadRequestException, Injectable } from "@nestjs/common";
 import { InjectRepository } from "@nestjs/typeorm";
 import { Repository } from "typeorm";
 import { ForksHistogramDto } from "../histogram/dtos/forks";
-import { DbForkGitHubEventsHistogram } from "./entities/fork_github_events_histogram";
+import { DbForkGitHubEventsHistogram, DbTopForkGitHubEventsHistogram } from "./entities/fork_github_events_histogram";
 
 @Injectable()
 export class ForkGithubEventsService {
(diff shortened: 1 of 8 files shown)


src/timescale/fork_github_events.service.ts
```diff
@@ -41,4 +41,22 @@ export class ForkGithubEventsService {
...
     return rawResults as DbForkGitHubEventsHistogram[];
   }
+
+  async genForkTopHistogram(): Promise<DbTopForkGitHubEventsHistogram[]> {
+    const queryBuilder = this.baseQueryBuilder();
+
+    queryBuilder
+      .select("repo_name")
+      .addSelect("time_bucket('1 day', event_time)", "bucket")
+      .addSelect("count(*)", "fork_count")
```
nickytonline, line 51: My SQL chops are rusty, but is using `*` still bad for performance?

jpmcb: Nah, this just counts the number rows we are selecting: aka, it counts the existence of returned rows. The alternative would be `count(repo_name)` which actually may have worse performance since you'd be instructing the query selector to find a specific row in the results to then count: using the `*` just says "count the returned rows" instead of "count the returned rows with column ... "

From the postgres docs:

> count ( * ) → bigint - Computes the number of input rows. 
> count ( "any" ) → bigint - Computes the number of input rows in which the input value is not null.

So `count(*)` is actually cheaper since it does not include that additional check for `"any" is not null`.

What you may be thinking of is `select *` which is less performant than selecting for specific columns since you'd be selecting _all_ columns in the given table.

nickytonline (approved):

brandonroberts (approved):

PR #642, opened 2024-03-12
fix: use skip/take for pagination with nested data for insights

brandonroberts:
## Description

- Fixes an issue when using pagination where the cumulative total records returns gets truncated when using joins across tables.
- One repository insight may have 10 repositories and 10 member associated with it and even if you have 11 total insights, it may only show 1 because the repos/members count against the pagination
- Using skip/take instead of offset/limit resolves this issue when querying data so all insights rows are returned before doing the joins and relational entities

- [ ] 🍕 Feature
- [x] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI
- [ ] 📦 Chore (Release)
- [ ] ⏩ Revert

## Added to documentation?

- [ ] 📜 README.md
- [ ] 📓 docs.opensauced.pizza
- [ ] 🍕 dev.to/opensauced
- [ ] 📕 storybook

### DIFF
--- a/src/workspace/workspace-insights.service.ts
+++ b/src/workspace/workspace-insights.service.ts
@@ -68,8 +68,8 @@ export class WorkspaceInsightsService {
       .leftJoinAndSelect("workspace_insights_insight.repos", "workspace_insights_insight_repos")
       .where("workspace_insights.workspace_id = :id", { id })
       .orderBy("workspace_insights_insight.updated_at", "DESC")
-      .offset(pageOptionsDto.skip)
-      .limit(pageOptionsDto.limit);
+      .skip(pageOptionsDto.skip)
+      .take(pageOptionsDto.limit);
 
     const itemCount = await queryBuilder.getCount();
     const entities: DbInsight[] = await queryBuilder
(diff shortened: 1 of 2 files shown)


src/workspace/workspace-insights.service.ts
```diff
@@ -68,8 +68,8 @@ export class WorkspaceInsightsService {
       .leftJoinAndSelect("workspace_insights_insight.repos", "workspace_insights_insight_repos")
       .where("workspace_insights.workspace_id = :id", { id })
       .orderBy("workspace_insights_insight.updated_at", "DESC")
-      .offset(pageOptionsDto.skip)
-      .limit(pageOptionsDto.limit);
+      .skip(pageOptionsDto.skip)
```
nickytonline, line 71: For reference, https://github.com/nestjsx/nestjs-typeorm-paginate/issues/576#issuecomment-936718238

nickytonline (approved):

src/workspace/workspace-insights.service.ts
jpmcb: > use skip / take instead of limit / offset when you do joins with selection.

There are _alot_ of places we need to fix this then. Example:

https://github.com/open-sauced/api/blob/7e86446487333657584c97832e2aab9f1479eb97/src/workspace/workspace-insights.service.ts#L71-L72

Issue #687, opened 2024-03-28
Feature: introduce `v2/workspaces/.../repos/prs` to get prs for repos in workspace


jpmcb:
### Type of feature

🍕 Feature

### Current behavior

Currently, to get a list of PRs for workspace repos, a client must aggregate the repos and then call `v2/repos/search`. This is problematic since the `v2/workspace/.../repos` endpoint is paginated and it would require iterating all pages to get all repos to then find the PRs for.

Not ideal!

Related to:
- https://github.com/open-sauced/app/pull/3048

### Suggested solution

Add a `v2/workspaces/.../repos/prs` endpoint that can paginate prs (and apply filters) to the workspace repo's PRs

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

### Contributing Docs

- [X] I agree to follow this project's Contribution Docs

PR #692, opened 2024-03-28
feat: Add `v2/workspaces/:id/repo/prs` endpoint for workspace repo PRs

jpmcb:
## Description

This implements a workspace repo PRs endpoint that can be used to get a filtered list of pull requests for a given workspace.

- [x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI
- [ ] 📦 Chore (Release)
- [ ] ⏩ Revert

Closes: https://github.com/open-sauced/api/issues/687

```sh
curl -X 'GET' \
  'http://localhost:3003/v2/workspaces/8d9f8ee1-b5dc-452c-b547-6cd2e375f1ec/repos/prs?page=1&limit=10&range=30&prev_days_start_date=0&repoIds=20580498%2C88664721%2C57939112&contributor=siyuanfoundation' \
  -H 'Authorization: Bearer <redacted>'
```

^ this gets prs for the repos: `20580498`, `88664721`, `57939112` where the contributor is `siyuanfoundation`.

This is more or less the same as `v2/prs/search` but requires `repoIds` (a comma delimited list of repo ids) and drops the topic, filter stuff. It does a pretty lazy check of the repos being in the workspace by querying for those repos in the `workspace_repos` table and making sure the lengths match: should be good for now but if we need to expand this feature, might require more validation.

## Tier (staff will fill in)

- [ ] Tier 1
- [ ] Tier 2
- [ ] Tier 3
- [x] Tier 4


src/workspace/workspace-repos.service.ts
```diff
@@ -69,6 +73,48 @@ export class WorkspaceReposService {
...
+    queryBuilder
+      .withDeleted()
+      .leftJoinAndSelect(
+        "workspace_repos.repo",
+        "workspace_repos_repo",
+        "workspace_repos.repo_id = workspace_repos_repo.id"
+      )
+      .where("workspace_repos.deleted_at IS NULL")
+      .andWhere("workspace_repos.workspace_id = :id", { id })
+      .andWhere("workspace_repos.repo_id IN (:...repoIds)", { repoIds })
```
nickytonline, line 105: Since this is optional filtering (passing in repoIds), should we only add this `.andWhere` if there are `repoIds` passed in?

jpmcb: In this context, the repo IDs are not optional. I figured that you'd always be providing repo IDs and that fit nicely into calling

```ts
this.pullRequestGithubEventsService.findAllWithFilters(pageOptionsDto);
```

which has no notion of the workspaces.

Would it make more sense to make repo IDs optional? In that case, if repo IDs is not provided, I'd need to go get all the IDs from the `workspace_repos` table. Up to you and how you'd like the API to be shaped!!

nickytonline: Maybe I understood the mission. 😅 Based on the PRs table protoype in https://github.com/open-sauced/app/issues/2934, since this is activity withing a workspace, I assume it would be listing all pull requests (paged) for all tracked repositories and from there if someone wanted to drill down to specific repositories, we'd have a dynamic filter that'd query the tracked repos in the workspace. cc: @isabensusan

jpmcb: ^ yes, that's correct from my understanding. My comments above are purely technical details. It just depends on how you want the API shaped: should the caller provide _all_ repoIDs to return at all times? Or provide the repoIDs to filter out?

The way this is implemented right now, only the repoIDs provided will be returned. I can easily flip it to be the inverse of that flow.

nickytonline: I think the inverse would make sense, so the whole kitchen sink when they go to the page, and let them filter repos as they see fit after that.

jpmcb: Cool cool 🚰 makes sense: sorry for the confusion here! Should have that turned around quickly

nickytonline (approved):
Thanks @jpmcb! 🚢

src/workspace/workspace-repos.service.ts
```diff
@@ -69,6 +73,56 @@ export class WorkspaceReposService {
...
+        "workspace_repos_repo",
+        "workspace_repos.repo_id = workspace_repos_repo.id"
+      )
+      .where("workspace_repos.deleted_at IS NULL")
+      .andWhere("workspace_repos.workspace_id = :id", { id });
+
+    if (pageOptionsDto.filterOutRepoIds) {
+      queryBuilder.andWhere("workspace_repos_repo.id NOT IN (:...repoIds)", {
+        repoIds: pageOptionsDto.filterOutRepoIds.split(","),
+      });
```
jpmcb, line 108: Now, when provided with the `filterOutRepoIds`, these will not be included when querying for PRs 🚀

brandonroberts (approved):

Issue #700, opened 2024-04-02
Feature: Add option for 12 months/365 days to range options


brandonroberts:
This is definitly possible and actually wouldn't be too much more expensive if we craft the queries right:

```sql
select * from pull_request_github_events
where lower(pr_author_login) = 'bdougie'
and event_time > now() - interval '12 months';
```

```
214 ms
150 rows
```

in fact, i _believe_ all it would take on the API side would be unlocking a different enum type for the calls to those endpoints (where currently it's `7 days, 30 days, 90 days`. Could expand that to `180 days, 360 days`)

_Originally posted by @jpmcb in https://github.com/open-sauced/engineering/issues/279#issuecomment-2030010947_

PR #701, opened 2024-04-02
feat: Support 6 month and 1 year ranges

jpmcb:
## Description

Adds support in the DTOs for:
- 6 month ranges
- 1 year ranges

Closes: https://github.com/open-sauced/api/issues/700

## Steps to QA

Went through all endpoints the use a range on `GET`: many don't seem to actually need it (since we just use a base DTO for all paging - maybe in the future we are more particular about which DTOs get used where).

I added an escape statement to a few spots where I saw some obvious performance problems: if we want bigger ranges with those requests, we'll need to rethink how those queries work.

The `/v2/histogram` endpoints work flawlessly and will be a great use case for surfacing time series data for these longer ranges. For example, the pushes histogram returns in less than 1 second for over 1 year of data on the `open-sauced/app` repo:

Screenshot 2024-04-02 at 2 58 51 PM

I'm certain we'll hit some performance edge cases where we'll need to rethink queries with longer ranges. We can raise issues as we run into those edges.

## Tier (staff will fill in)

- [ ] Tier 1
- [ ] Tier 2
- [ ] Tier 3
- [x] Tier 4


brandonroberts (approved):

nickytonline (approved):
:shipit:

PR #737, opened 2024-04-22
feat: add devstats to owner/repo endpoint

zeucapua:

Closes #729 

## Tier (staff will fill in)

- [ ] Tier 1
- [ ] Tier 2
- [ ] Tier 3
- [x] Tier 4

### DIFF
--- a/src/repo/repo.service.ts
+++ b/src/repo/repo.service.ts
@@ -85,7 +85,7 @@ export class RepoService {
     return item;
   }
 
-  async findOneByOwnerAndRepo(owner: string, repo: string): Promise<DbRepoWithStats> {
+  async findOneByOwnerAndRepo(owner: string, repo: string, range?: number): Promise<DbRepoWithStats> {
     const queryBuilder = this.baseQueryBuilder();
 
     queryBuilder.where("LOWER(repo.full_name) = :name", { name: `${owner}/${repo}`.toLowerCase() });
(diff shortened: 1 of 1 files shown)


src/repo/repo.service.ts
```diff
@@ -96,7 +96,35 @@ export class RepoService {
       throw new NotFoundException(`Repository not found: ${owner}/${repo}`);
     }
 
-    return item;
+    const prStats = await this.pullRequestGithubEventsService.findPrStatsByRepo(
```
jpmcb, line 99: Just needs a `npm run format` run to clean this up 🧹

src/repo/repo.service.ts
```diff
@@ -96,7 +96,35 @@ export class RepoService {
...
+      item.full_name,
+      range || 30,
```
jpmcb, line 101: I recently cleaned up some of these:

```
Prefer using nullish coalescing operator (`??`)
instead of a logical or (`||`), as it is a safer operator
```

I _think_ this was a linting rule set up early early on. Not sure if it's really at all relevant or needed.

src/repo/repo.service.ts
```diff
@@ -96,7 +96,35 @@ export class RepoService {
...
+      closed_prs_count: prStats.closed_prs,
+      pr_velocity_count: prStats.pr_velocity,
+      fork_velocity: forksVelocity,
+      activity_ratio: activityRatio,
+      contributor_confidence: confidence,
+      health: activityRatio,
+      last_pushed_at: pushDates.push_date,
+      last_main_pushed_at: pushDates.main_push_date,
+    } as DbRepoWithStats;
+    // return item;
```
jpmcb, line 127: Unnecessary comment

src/repo/repo.service.ts
```diff
@@ -85,7 +85,7 @@ export class RepoService {
     return item;
   }
 
-  async findOneByOwnerAndRepo(owner: string, repo: string): Promise<DbRepoWithStats> {
+  async findOneByOwnerAndRepo(owner: string, repo: string, range?: number): Promise<DbRepoWithStats> {
```
jpmcb, line 88: Could just set a default here instead of doing several `||` checks on the optional (defaults should also be optional)

```suggestion
  async findOneByOwnerAndRepo(owner: string, repo: string, range: number = 30): Promise<DbRepoWithStats> {
```

src/repo/repo.service.ts
```diff
@@ -96,7 +96,30 @@ export class RepoService {
...
+      open_prs_count: prStats.open_prs,
+      merged_prs_count: prStats.accepted_prs,
+      spam_prs_count: prStats.spam_prs,
+      draft_prs_count: prStats.draft_prs,
+      closed_prs_count: prStats.closed_prs,
+      pr_velocity_count: prStats.pr_velocity,
+      fork_velocity: forksVelocity,
+      activity_ratio: activityRatio,
+      contributor_confidence: confidence,
+      health: activityRatio,
```
nickytonline, line 119: This is temporary correct @jpmcb? I believe we do this elsewhere as well, but we only consume the `activity_ratio` in the frontend, and never `health`. Should this property be omitted for the moment?

jpmcb: Yeah, we do it in the prior-art in repo search:

https://github.com/open-sauced/api/blob/e84030eddb8e952e35cd164d8c39635403335030/src/repo/repo.service.ts#L202-L217

So, for 1 to 1 parity, probably can include this as Zeu has it 👍🏼 in the future, we'll expand health to include other metrics

jpmcb (approved):

brandonroberts (approved):

PR #742, opened 2024-04-24
fix: expose `range` query param for `/repo/<org>/<repo>`

### LINKED ISSUE #740: Bug: expose `range` parameter for `/repo/<org>/<repo>` endpoint
### Describe the bug

While the service for `findOneByOwnerAndRepo` has a parameter for range to be used to get the repository's devstats, there's no way to set it via the controller, thus always resulting to getting stats for the default of 30 days.

### Steps to reproduce

1. Use the `/repo/<org>/<repo>` endpoint and look for `open_prs_count`, `merged_prs_count`, and `pr_velocity_count`
2. Add a `range` query param to the URL.
3. See that it doesn't change the stats listed above.

### Do you have any images or screen captures?

_No response_

### Code of Conduct

### Contributing Docs

- [ ] I agree to follow this project's Contribution Docs

zeucapua:
## Description

Implements a new `RepoRangeOnlyOptionDto` parameter to the `/repo/<org>/<repo>` endpoint controller to surface the `range` option. This is so the devstats given for each repository reflects the correct data.

Closes #740 

## Tier (staff will fill in)

- [ ] Tier 1
- [ ] Tier 2
- [ ] Tier 3
- [x] Tier 4


brandonroberts (approved):
LGTM

src/repo/repo.service.ts
```diff
@@ -371,7 +378,7 @@ export class RepoService {
       if (repoId) {
         repo = await this.findOneById(repoId);
       } else if (repoOwner && repoName) {
-        repo = await this.findOneByOwnerAndRepo(repoOwner, repoName);
+        repo = await this.findOneByOwnerAndRepo(repoOwner, repoName, rangeOption?.range);
```
nickytonline, line 381: Should this not default to 30?

nickytonline (commented):
:shipit:

src/repo/repo.service.ts
zeucapua: Since `rangeOption` is optional, should this parameter be `rangeOption?.range! || 30` to ensure that there's a default?

zeucapua: The `rangeOption` DTO does have a default of 30, is there something I'm missing in implementing the default?

src/repo/repo.service.ts
```diff
@@ -373,12 +373,12 @@ export class RepoService {
     }
 
     let repo;
-
+    const range = rangeOption?.range! || 30;
```
nickytonline: This will probably work better.

```suggestion
const range = rangeOption?.range ?? 30;
```

nickytonline (approved):
🚢

PR #746, opened 2024-04-25
fix: remove default value from exceeds_ugprade_limit field on workspace

brandonroberts:
## Description

- Removes `default: false` field from entity
- It causes TypeORM to use it in the `insert` query even if `insert: false` is set on the field

## Tier (staff will fill in)

- [ ] Tier 1
- [ ] Tier 2
- [ ] Tier 3
- [x] Tier 4


nickytonline (approved):
LGTM as we just paired on this.

:shipit:

bdougie (approved):

PR #756, opened 2024-05-01
feat: Refactor user-list most-active-contributors query

jpmcb:
## Description

Offloads much of the data crunching onto the API form getting most active contributors commits, prs-created, pr-reviews, etc. etc.

Instead, it makes many small concurrent queries to the database to be aggregated when the queries complete.

This carries abit of a risk of offloading huge amounts of data into raw memory which could throttle the API (instead of throttling the database). But, in theory, these queries should complete relatively quickly and we can rely on Nest JS / garbage collection to do it's job.

Related to https://github.com/open-sauced/api/issues/695 and user-list query optimizations.

## Steps to QA

Call the `stats/most-active-contributors` endpoint like:

```
v2/lists/8c329a5a-1c4e-4108-aca8-6d0aa1915cbb/stats/most-active-contributors
```

and see data get returned. For really massive lists, this still might take a few seconds. But we can be assured the database won't get completely pegged.

This list, with over 450 contributors, took about 20 seconds:

```
❯ time curl -X 'GET' \
  'http://localhost:3003/v2/lists/8c329a5a-1c4e-4108-aca8-6d0aa1915cbb/stats/most-active-contributors?page=1&limit=50&range=30&prev_days_start_date=0' -H 'accept: application/json'

0.00s user 0.01s system 0% cpu 19.223 total
```

This is a massive improvement over completely throttling out the timescale database.

## Tier (staff will fill in)

- [ ] Tier 1
- [ ] Tier 2
- [ ] Tier 3
- [x] Tier 4


jpmcb:
This is ready for others to look at, just need to do some additional validation / testing to get it to the finish line!

brandonroberts (approved):

nickytonline (approved):

PR #760, opened 2024-05-02
fix: Refactors contributions-by-project endpoint

jpmcb:
## Description

This refactors the `contributions-by-project` query used in the contributors insights.

Now, the endpoint pieces together data for each contributor instead of attempting to do one massive (often very slow) query that pegs the database. For a normal sized list, this should return withing a second. And for a massive list, this returns within about +20 seconds.

Related to: https://github.com/open-sauced/api/issues/695

## Steps to QA

A list of about 20 contributors:

```
time curl -X 'GET' \
  'http://localhost:3003/v2/lists/39cd0fde-f778-401a-94d7-683d3b732808/stats/contributions-by-project?range=30' \
  -H 'accept: application/json'

0.00s user 0.01s system 0% cpu 1.370 total
```

A huge list of over 450+ contributors:

```
time curl -X 'GET' \
  'http://localhost:3003/v2/lists/8c329a5a-1c4e-4108-aca8-6d0aa1915cbb/stats/contributions-by-project?range=30' \
  -H 'accept: application/json'

0.00s user 0.01s system 0% cpu 37.340 total
```

## Tier (staff will fill in)

- [ ] Tier 1
- [ ] Tier 2
- [ ] Tier 3
- [x] Tier 4


src/timescale/contrib-stats.service.ts
```diff
@@ -88,4 +93,138 @@ export class ContributorDevstatsService {
       total_contributions: commits + prs_created + prs_reviewed + issues_created,
     });
   }
+
+  async findAllContributionsByProject(
+    pageOptionsDto: ContributionsByProjectDto,
+    users: string[]
+  ): Promise<DbContributionsProjects[]> {
+    const aggregatedStats: Record<string, DbContributionsProjects> = {};
```
jpmcb, line 101: The TLDR of this PR is we now build a hashmap (through TypeScript's `Record` typing) where the keys are the `repo_name` for each of the individual stat points for the user. We then aggregate it and all together to return the final resulting list of `DbContributionsProjects`

src/timescale/commit_comment_github_events.service.ts
```diff
@@ -19,12 +20,12 @@ import { ContributorStatsTypeEnum } from "./dtos/most-active-contrib.dto";
 @Injectable()
 export class CommitCommentGithubEventsService {
   constructor(
-    @InjectRepository(DbCommitCommentGitHubEventsHistogram, "TimescaleConnection")
-    private commitCommentGitHubEventsHistogramRepository: Repository<DbCommitCommentGitHubEventsHistogram>
+    @InjectRepository(DbCommitCommentGitHubEvents, "TimescaleConnection")
+    private commitCommentGitHubEventsRepository: Repository<DbCommitCommentGitHubEvents>
```
jpmcb, line 24: I renamed a few of the Timescale classes injected repositories: this is more or less naming cleanup since it doesn't really make sense to use the "Histogram" repository as a general foothold for TypeORM. You'll also notice that I cleanup the `baseQueryBuilder` for a few classes: this is also in service of the base query builder not returning an agnostic `manager` (which ignores the TypeORM table for that repository).

src/timescale/contrib-stats.service.ts
nickytonline: I wonder if a [WeakMap](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/WeakMap) would be beneficial here?

nickytonline: I wonder if a [WeakMap](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/WeakMap) would be beneficial here?

bdougie (approved):
🚀🚀🚀

src/timescale/contrib-stats.service.ts
jpmcb: A a typed `Record` and `WeakMap` look almost the same. Although, some of the discourse online seems to suggest that:

1. Prefer `Record` since you don't have to instantiate it (you just set it to an empty object `= {}` vs. `const wm = new WeakMap();`
2. Prefer `Record` since the record's data values are strongly typed while the `Map` and `WeakMap` may accept other datatypes as values.

brandonroberts (approved):
Tested locally and the graph data loaded successfully and faster

PR #787, opened 2024-05-11
feat: Skunkworks - Star-search send "functionCall" events to clients

jpmcb:
## Description

Adds the ability for the star-search observable to send the "event" type down the pipeline to the client alongside the json schema payload that was called with the function.

Relates to: https://github.com/open-sauced/app/issues/3353

## Steps to QA

Ask a question and notice the events now have a "event" metadata alongside the data for the json:

```
❯ ./hack/call-sse.sh

event: function_call
id: 1
data: {"name":"searchPrsByAuthor","arguments":"{\"question\": \"What recent pull requests has @brandonroberts been involved in?\", \"author\": \"brandonroberts\"}"}

event: function_call
id: 2
data: {"name":"searchIssuesByAuthor","arguments":"{\"question\": \"What recent issues has @brandonroberts been involved in?\", \"author\": \"brandonroberts\"}"}

...

event: content
id: 5
data: Brandon

event: content
id: 6
data:  Roberts

event: content
id: 7
data:  has

event: content
id: 8
data:  been

...
```

The really tricky thing here will be the client side validation of these non-deterministic events. I've seen it have to re-call a tool since it didn't provide a required argument initially. We'll likely need some very strongly typed schemas on the client (like we have in the API here) to ensure reliability between what the model is calling and what gets sent down to the client.

## Tier (staff will fill in)

- [ ] Tier 1
- [ ] Tier 2
- [ ] Tier 3
- [x] Tier 4


src/openai-wrapped/openai-wrapped.service.ts
```diff
@@ -8,6 +8,8 @@ import { JSONSchema } from "openai/lib/jsonschema";
 import { ZodSchema } from "zod";
 import { zodToJsonSchema } from "zod-to-json-schema";
 
+type AsyncOrSyncFunction<T extends object, R> = (args: T) => Promise<R> | R;
```
jpmcb, line 11: This was a small change to hoist out the function type for the zod generated schema: in the future, we can use this to also call non-async functions that don't return anything (like when we just want to render components).

jpmcb:
Pushed a commit that includes a tool for signaling to render the lottery factor component. This appears in the SSE as:

```
What is the lottery factor of kubernetes/kubernetes?
```

```
event: function_call
id: 1
data: {"name":"renderLottoFactor","arguments":"{\"repoName\":\"kubernetes/kubernetes\"}"}
```

which is the OpenAI `ChatCompletionMessage.FunctionCall` type in the data for the `renderLottoFactor` function call.

```ts
export const RenderLottoFactorParams = z.object({
  repoName: z.string(),
});
export type RenderLottoFactorParams = z.infer<typeof RenderLottoFactorParams>;
```

This still produces some text / markdown content:

> The "Lottery Factor" for the repository `kubernetes/kubernetes` is low. This indicates that the risk of the project\'s future being jeopardized by the sudden unavailability of its most active contributors is minimal. The distribution of contributions among multiple active members helps ensure the project\'s stability and continuity.

and in the future we could get the LLM to not say anything after it's called this tool.

---

cc @nickytonline, let me know if you think this needs anything else from the client side!

brandonroberts (approved):
LGTM 🦨 🧑‍🏭

src/star-search/star-search.controller.ts
```diff
@@ -22,18 +31,25 @@ export class StarSearchController {
...
         .on("content", (delta) => {
           observer.next({
+            type: StarSearchObservableEventTypeEnum.content,
             data: delta,
           });
         })
         .on("message", (msg) => console.log("msg", msg))
-        .on("functionCall", (functionCall) => console.log("functionCall", functionCall))
+        .on("functionCall", (functionCall: ChatCompletionMessage.FunctionCall) => {
+          console.log("functionCall", functionCall);
```
brandonroberts, line 47: Are we leaving this in?

jpmcb: Yeah, for now. It's been really useful to understand how the LLM chooses to use which tools. In the future, we probably want to find a more intelligent way to log to a logging service the flow of what the LLM chooses.

brandonroberts: Cool, figured so 👍

nickytonline (approved):
Based on our convos, LGTM.

:shipit:

PR #804, opened 2024-05-16
fix: validate the repository before requesting the lottery factor to be rendered

brandonroberts:
## Description

- Adds a check to make sure the provided repositories are valid before returning the requested lottery factor.
- Discussed on Slack about invalid repos returning data for StarSearch

## Tier (staff will fill in)

- [ ] Tier 1
- [ ] Tier 2
- [ ] Tier 3
- [x] Tier 4


nickytonline (approved):
:shipit:

jpmcb (changes requested):
Played around with this and was getting:

Screenshot 2024-05-16 at 3 51 51 PM

where it just wouldn't respond with anything. Looking at my local logs, looks like we need to handle the try / catch for for the actual tool function (and `await` the promise to get the error) since it doesn't seem the function calling loop the manager agent enters will handle thrown errors: seems it just throws out of the stack and gets caught by the nest controller.

---

2 suggestions that i tried out:

```diff
diff --git a/src/repo/repo.service.ts b/src/repo/repo.service.ts
index 84b13cea..b46114bc 100644
--- a/src/repo/repo.service.ts
+++ b/src/repo/repo.service.ts
@@ -442,7 +442,11 @@ export class RepoService {
       });
     } catch (error: unknown) {
       console.error(error);
-      throw new BadRequestException("Error fetching repo:", `${owner}/${repo}`);
+      if (error instanceof Error) {
+        throw new BadRequestException("Error fetching repo:", `${owner}/${repo} - ${error.message}`);
+      } else {
+        throw new BadRequestException("Error fetching repo:", `${owner}/${repo} - Unknown error`);
+      }
     }

     const parts = octoResponse.data.full_name.split("/");
diff --git a/src/star-search/star-search-tools.service.ts b/src/star-search/star-search-tools.service.ts
index 6ba45613..26446c15 100644
--- a/src/star-search/star-search-tools.service.ts
+++ b/src/star-search/star-search-tools.service.ts
@@ -29,7 +29,15 @@ export class StarSearchToolsService {
    */

   async renderLottoFactor({ repoName }: RenderLottoFactorParams) {
-    return this.repoService.findLottoFactor({ repos: repoName });
+    try {
+      const result = await this.repoService.findLottoFactor({ repos: repoName });
+
+      return result;
+    } catch (error: unknown) {
+      if (error instanceof Error) {
+        return `error getting lottery factor data: ${error.message}`;
+      }
+    }
   }

   runTools(question: string): ChatCompletionStreamingRunner {
```

1. Modify the error messaging slightly where Octokit throws out the error (so that we actually see "404" from Ocotkit)
2. Do a try / await / catch flow in the `renderLottoFactor` tool function for the AI to handle the error case.

This seemed to work well when I applied the ^ diff:

Screenshot 2024-05-16 at 3 54 46 PM

brandonroberts:
Thanks, patches applied.

jpmcb (approved):

PR #818, opened 2024-05-18
feat: Introduces a strongly typed, defined data object for StarSearch

jpmcb:
## Description

This introduces a strongly typed, well defined data object for the `data:` member of the stream for StarSearch. This should help us get around _alot_ of the pain we've been having with rendering on the frontend. This data object looks like:

This is _heavily_ inspired by how ChatGPT works and the shape of the streamed payloads you get back from their `https://chatgpt.com/backend-api/conversation` endpoint.

---

```json5
{
  // Some ID: in the future, this would likely be a uuid of the message flowing
  // through that we could trace in case there are individual issues on individual messages
  "id": "123-abc",

  // At this point, this will always be "manager" since the StarSearch manager agent is the
  // only one sending content
  "author": "manager",
  "iso_time": "2024-05-18T22:33:55.224Z",
  "content": {
     // will be either "content" or "function_call"
    "type": "content",

    // this is the actual content. At this time, it'll be either
    // - the _combined_ string message from the agent
    // - the stringified JSON from the function_call 
    "parts": [
      "The Lottery Factor for the `kubernetes/kubernetes` repository is **low**. This suggests that the"
    ]
  },

  // either "in_progress" or "done"
  "status": "in_progress",

  // This also gives us the capability to send a error object per message if we'd like.
  "error": null
}
```

This all combined ends up looking like:

```
id: 1
data: {"id":"789-xyz","author":"manager","iso_time":"2024-05-18T22:33:51.933Z","content":{"type":"function_call","parts":["{\"name\":\"renderLottoFactor\",\"arguments\":\"{\\\"repoName\\\":\\\"kubernetes/kubernetes\\\"}\"}"]},"status":"in_progress","error":null}

id: 2
data: {"id":"123-abc","author":"manager","iso_time":"2024-05-18T22:33:55.034Z","content":{"type":"content","parts":["The"]},"status":"in_progress","error":null}

id: 3
data: {"id":"123-abc","author":"manager","iso_time":"2024-05-18T22:33:55.034Z","content":{"type":"content","parts":["The Lottery"]},"status":"in_progress","error":null}

id: 4
data: {"id":"123-abc","author":"manager","iso_time":"2024-05-18T22:33:55.035Z","content":{"type":"content","parts":["The Lottery Factor"]},"status":"in_progress","error":null}

id: 5
data: {"id":"123-abc","author":"manager","iso_time":"2024-05-18T22:33:55.035Z","content":{"type":"content","parts":["The Lottery Factor for"]},"status":"in_progress","error":null}
```

Where each stream data chunk appends to the `parts` string for the total content from the LLM.

## Tier (staff will fill in)

- [ ] Tier 1
- [ ] Tier 2
- [ ] Tier 3
- [x] Tier 4


src/star-search/star-search.controller.ts
```diff
@@ -49,20 +42,43 @@ export class StarSearchController {
 
     const stream = this.starSearchToolsService.runTools(preProcessResults.prompt);
 
+    let message = "";
+
     return new Observable<StarSearchObservable>((observer) => {
       stream
         .on("content", (delta) => {
+          message += delta;
```
nickytonline, line 50: so instead of sending delta's chunk by chunk it's still sending chunks, but the chunk is accumulative so that the last chunk is the full response?

nickytonline (commented):
:shipit: but hold off on merging until the app

src/star-search/star-search.controller.ts
jpmcb: Exactly: I have the PR in the app that sets the content from this payload each time instead of appending! So, works basically the same with both these combined, just more complete (with spacing and everything included)

nickytonline (approved):
🚢

src/star-search/star-search.controller.ts
```diff
@@ -49,20 +42,43 @@ export class StarSearchController {
...
+
           observer.next({
-            type: StarSearchObservableEventTypeEnum.content,
-            data: delta,
+            data: {
+              id: "123-abc",
```
brandonroberts, line 54: Just a note. I thought this was test data, but its for separating the events sent back 👍

brandonroberts (approved):

src/star-search/star-search.controller.ts
jpmcb: > its for separating the events sent back

Yeah, this is sort of a punt since the client doesn't really inspect the "id" of any one given message coming in. But in the future, these could be real unique identifiers and we could do something more interesting.

jpmcb:
Holding off on merging this until https://github.com/open-sauced/app/pull/3424 is ready to go in too

bdougie (approved):
Your "i love zod" tweet makes a lot of sense now.

Issue #648, opened 2024-03-13
Bug: add missing pipe validators


jpmcb:
### Describe the bug

There are a few places in the Nest controllers where we are missing pipe validators which can cause 500s. For example, in the workspace delete member controller:

https://github.com/open-sauced/api/blob/ecb7657374ece5181863b235f45df69ab86ccedc/src/workspace/workspace-members.controller.ts#L102-L106

providing the `id: number` as a _string_ will cause a 500 since there's no validator in this flow.

I'm sure there are a few places this is happening that we'll want to add validators in.

### Steps to reproduce

1. Attempt to delete workspace member with a malformed body:

```json
{
  "members": [
    {
      "id": "woofwoof"
    }
  ]
}
```

2. notice 500 is thrown

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

### Contributing Docs

- [X] I agree to follow this project's Contribution Docs

takanome-dev:
Hey @jpmcb, wouldn't validating each `id` in that array with `@ValidateNested({each: true})` fix that issue (taking the example you provided)?

Would love to help with this but cannot get my server up and running 😑

I'm having two issues currently:
```zsh
[Nest] 90442  - 03/15/2024, 1:39:20 PM   ERROR [TypeOrmModule] Unable to connect to the database (TimescaleConnection). Retrying (4)...
error: role "tsdbadmin" does not exist
```
Don't know yet how to add that role in my timescale db
```zsh
[Nest] 90442  - 03/15/2024, 1:39:20 PM   ERROR [TypeOrmModule] Unable to connect to the database (ApiConnection). Retrying (4)...
error: SSL connection is required
```
I did export my ssl cert from supabase and put it in my env file (replacing new lines with `\n`) but 😑

jpmcb:
Hi @takanome-dev - thanks for the help here: you won't be able to get the API up and running without an active Timescale connection which is our service for sourcing GitHub events. I'll take this on 👍🏼

PR #819, opened 2024-05-20
feat: Pipe validators for UUIDs

jpmcb:
## Description

Fixes missing pipe validators for UUIDs. Otherwise, we'd get 500s since the raw strings would get used as UUIDs.

Related to #648 

Followup to: https://github.com/open-sauced/app/pull/3419 which was seeing 500s when `new` was passed in as the workspace ID. This now should throw a 400 bad request instead of having an internal server error.

Screenshot 2024-05-20 at 12 36 20 PM

## Tier (staff will fill in)

- [ ] Tier 1
- [ ] Tier 2
- [ ] Tier 3
- [x] Tier 4


brandonroberts (approved):
And its built-in! 👏

